PR #504, opened 2015-01-28
Update mime.types to the latest version.

XhmikosR:
https://github.com/h5bp/server-configs-nginx


jmervine:
Note to self, when releasing...

```
make nginx.conf nginx/reload
```

PR #509, opened 2015-02-25
Updates

XhmikosR:
@jmervine: I'm having issues installing the dependencies with node.js 0.12 that is why I decided to add it in Travis.

Let me know if you need me to change anything before you merge this.


jmervine:
I wouldn't expect 0.12 to work yet. In fact most applications that I work on, near or around are sticking to 0.10.x for now due to compatibility issues with dependent modules. I would like to upgrade to 0.10.36 sooner than later, but don't have any short term plans to upgrade to 0.12.x.

J

.travis.yml
```diff
@@ -1,4 +1,7 @@
 language: node_js
 node_js:
   - "0.10"
+  - "0.12"
 script: make travis
```
jmervine: This should now be `node make travis`, no?

jmervine: Actually, `make travis` should still exist, so perhaps that's not the issue.

XhmikosR: Well, I agree, but last time you said you wanted to keep the makefile. Personally, I'm in favor of make.js since it works everywhere.

XhmikosR:
Updated.

XhmikosR:
Travis still fails for this PR for node.js 0.12, but the exit code is wrongly 0 so it's marked as passed...

jmervine:
Okay, I'll do a bit more reseah.

jmervine:
With these commits, the 0.12 issues should be passing on travis, although I'll admit, I'm downgrading the test around an issue that's not reproducible outside of the tests. Additionally, I fixed the issue with make.js and not returning the correct exit status. 

Merge develop and I'll be this PR will pass for real'z. :)

XhmikosR:
Rebased :)

jmervine:
Sweet, looks like everything passed. You ready for me to merge this?

XhmikosR:
I'd say it's ready for now. We do need to fix the way we run validator and bootlint eventually but that's for another time.

jmervine:
Okay, sounds good. I’m too tired to release tonight. I’ll do it tomorrow morning when I’m fresh… just in case. =)

XhmikosR:
Updated. I can no longer reproduce the issue I had with forever 0.14.1 so I updated that too.

jmervine:
Bah, totally forgot about this. Been really busy. I'll look at it tonight.

.travis.yml
jmervine: I understand your view on that. I hesitate because in 9000 other things I do, it's "make ..." so it's basically muscle memory for me. For my own sanity I try to keep things consistent.

XhmikosR: In theory doing make should work on nix if the proper shebang is present,
shouldn't it?

jmervine:
@XhmikosR Please verify these changes on http://dev.bootstrapcdn.com and if all is well, let me know. They look good to me. I'll release to prod tomorrow afternoon some time.

.travis.yml
jmervine: Bang and perms, but it would still need to be './make'.

XhmikosR:
Everything seems to be fine @jmervine. BTW, http://www.bootstrapcdn.com/ returns 502.

jmervine:
Just saw this, restart it.

PR #523, opened 2015-03-16
[CODE REVIEW] Refactor for port based sharding.

jmervine:
cc @XhmikosR @jdorfman @jimaek 

This branch is being developed on to support migration the current single VM based host, to deployment on a larger box with multiple applications. For stability and performance, I want to run three separate instances of the application. As part of this I'm switching from using forever and `cluster` to simply using PM2. Additionally, I'm pairing down the Nginx configs to support being included in a larger configuration, which supports multiple vhost.

Lastly, I've removed everything from Makefile with the exception of a simple pass through to make.js.

Known remaining cleanup items:

* remove all references to legacy nginx configs. (i.e. `.nginx.conf`)
* refactor test runners to no longer required `forever`, thus completely removing it as a requirement

Am I missing anything else?


jmervine:
Thought I was going to make things prettier by squashing the commit instead of merging it, but it appears to include a bunch of files that haven't actually changed. Note:

```
jmervine@mervine bootstrap-cdn $ git status
On branch shard
Your branch is up-to-date with 'origin/shard'.
nothing to commit, working directory clean
jmervine@mervine bootstrap-cdn $ git diff --name-only develop
Makefile
bootstrapcdn.conf
bootstrapcdn.json
config/_config.yml
make.js
package.json
```

XhmikosR:
I'm pretty sure this branch isn't rebased properly...

jmervine:
@XhmikosR I agree, let me roll back this commit and do a standard merge.

XhmikosR:
You only need https://github.com/MaxCDN/bootstrap-cdn/commit/8211604556c4be6dc62d682e1a67aa278cb2df31 The other patch is redundant.

Just do `git rebase -i origin/master` and skip it.

jmervine:
@XhmikosR Okay, that looks better, at least diff-wise. While I actively use git every day, I feel like I need to take a class or something on rebase, it never works the way I would expect. :-P

XhmikosR:
What you did isn't a rebase still...

Just check my previous comment and skip the unneeded patches.

jmervine:
This is no longer valid.

PR #531, opened 2015-03-31
Use HTTPS where possible

mathiasbynens:
https://github.com/konklone/cdns-to-https#background

Closes #530.

Issue #530, opened 2015-03-31
Switch from "//" to "https://"?


jmervine:
See: https://twitter.com/mathias/status/582805207270166528

mathiasbynens:
I’m happy to submit a PR for this, so it’s easier to review.

mathiasbynens:
Done: #531.

jmervine:
@mathiasbynens Thanks for your contribution. I'm opening a dialog with @jdorfman and I'll get back to you on this soon.

jdorfman:
@jmervine This should be merged after reading www.paulirish.com/2010/the-protocol-relative-url/

jmervine:
Okay, I'm going to set aside some time to look at it. Need to get the tests passing. I'll probably not merge, but rather create an integration branch.

PR #534, opened 2015-04-17
HTTPS Update

jmervine:
Integration and updates from #531 (reference #530).

Changes:
* Tweets should be preserved as is.
* Content will not be updated, that's reserved for the original maintainers.


jmervine:
cc @jdorfman @mathiasbynens

jmervine:
Tests continue to fail b/c the "start" task was removed as part of the heroku migration. Certain tests require a running app. This needs to be addressed before moving forward. I'll try and work on it this weekend.

PR #535, opened 2015-04-17
Commenting out extras tests for now.

jmervine:
Required for tests to pass  in current state.


jmervine:
Tests continue to fail b/c the "start" task was removed as part of the heroku migration. Certain tests require a running app. This needs to be addressed before moving forward. I'll try and work on it this weekend.

PR #548, opened 2015-06-21
update bootswatch to 3.3.5

thomaspark:


jmervine:
Thanks for the PR! Restarted CI, as at a glance they should have passed, but occasionally travis times out.

PR #550, opened 2015-07-01
Syncing AWS S3 file store with git repo

jmervine:
cc @jdorfman

> ### Foreward
>
>A lot of the files on AWS S3 are very old and were uploaded well before the bootstrap-cdn current implementation (aka current git repo). As such, the process of syncing AWS S3 to github is a bit hairy, hence the detailed description, notes and multiple verbose commits below.

## Reasoning

1. We will be moving of AWS S3 as the CDN origin host. Why?
    - AWS S3 incurs additional complexity in having to ensure that files are synced regularly and properly.
    - AWS S3 limits control over file mime types. 
    - AWS S3 incurs additional, unnecessary cost. 
2. We feel that github should be in sync and authoritative.

## Method

1. Downloaded entire AWS S3 bucket using "awscli"

    ```
    aws s3 sync s3://bootstrap-cdn/public ./bootstrap-cdn-s3-files
    ```
2. Removed all resulting directories which where/are symlinks in github.
3. Copied files from `./bootstrap-cdn-s3-files` to `bootstrap-cdn/public` (git clone)

    ```
    cd ./bootstrap-cdn-s3-files
    rsync -r -v . /path/to/bootstrap-cdn/public
    ```
4. Using `git status` ensured that no symlinks were clobbered.
5. Using `git status`, selectively added newly introduced files.
6. Using `git diff`, reverted (via `git checkout path/to/file`) any files that clobbered known valid content from repo.
7. Adding remaining changes from S3 under the assumption that AWS S3 content is authoritative.

## Notes

1. Regarding `public/bootswatch/3.2.0` changes. I don't fully trust either source.
2. Regarding `public/html5shiv/*`, I'm fairly certain those files are no longer needed, but wanted to leave that to you.
3. Regarding `public/js/jquery-1.9.1.min.js`. I don't think this should be here at all, but didn't want to delete it without confirmation.
4. Regarding 'public/images/reports`, I have no clue why these changed, or what they're for. I'm pretty sure we should remove these from bootstrap. If they're needed for something else, we should move them to their own origin host/cdn location.
5. Regarding `public/stylesheets/{bootstrap-cdn,theme.bootstrap,docs}.css`, I'm pretty sure these can be deleted, but wanted verification.


jmervine:
@jdorfman please cc anyone you would like to have review this. I'll leave it to you to merge. Cheers!

jdorfman:
@jmervine great job.

Response to your **Notes**:

* 1: good call
* 2: **must be present**, legacy installations depend on it.
* 3: can be removed, legacy bootstrapcdn.com site file
* 4: can be removed, legacy bootstrapcdn.com site reporting icons
* 5: can be removed, legacy bootstrapcdn.com site files

jmervine:
@jdorfman removed mentioned files

jdorfman:
@jmervine looks good to me. Please deploy this to prod and I will update the origin url Monday.

jmervine:
Reopening to merge.

PR #559, opened 2015-07-24
Master

jdorfman:


jdorfman:
thanks great job (as always) @jmervine

PR #565, opened 2015-08-18
Added more tweets

jimaek:


jmervine:
Should we clean out some of the older ones, or do we want to just keep adding?

jdorfman:
@jmervine keep the old ones for now.  merging

jdorfman:
thank you @jimaek

PR #569, opened 2015-08-20
Adding bootstrap4 alpha support

jmervine:
@jdorfman 

Viewable here: https://bscdnpre.herokuapp.com/

> Note: maxcdn url won't work until until this is released to production.


jdorfman:
when does this go live?

PR #572, opened 2015-08-21
Releasing 4.0.0-alpha tab.

jmervine:
@jdorfman :-P

I'll merge this in the morning, before doing so, can you confirm that the identity here matches what's on the origin/cdn host(s). Given that we updated them separately, I'd be nice to be be sure. Cheers!


jmervine:
Verified the files were identical. Releasing.

PR #581, opened 2015-09-16
jd-integrations

jdorfman:
Removed space from 'Bootstrap CDN', added integrations (route, images, jade files), updated make.js & package.json with restart function


jmervine:
@jdorfman I'm not loving the way the submit integrations button is aligning down at the bottom, see: http://dev.bootstrapcdn.com/integrations/. I would probably put a blub about what these are and how to submit a new one, using similar formatting that we use in the Bootstrap4 Alpha page (http://dev.bootstrapcdn.com/alpha/)

jdorfman:
Good point.  What about the `restart` function?

jmervine:
I won't use it, nor will it be used by heroku, but if it helps you... it's all good.

jdorfman:
@jmervine yeah it helps me =)

PR #582, opened 2015-09-16
added radix & freight, removed button

jdorfman:
Added two more integrations. Removed the "Add an Integration" button from `views/integrations.jade`


jmervine:
@jdorfman I think travis failed because you merged the PR before it completed. We should let these run completely. Also, delete branches after merging to keep things clean, please. ;)

jdorfman:
@jmervine shit!  Sorry about that.  I agree I will start removing old branches.  Please feel free to remove any that are obsolete.

PR #597, opened 2015-11-21
Refactored existing tests, added a ton, fixed false positives, etc.

jmervine:
@jdorfman I took the previous headers tests, refactored and including tests against SRI as well. These all hit the CDN and therefore are true integration tests. I would actually like to run these regularly as monitoring (note to self). 

These can be merged to develop whenever and go out with the next release.

#### Update
This branch basically turned in to a total refactor of all tests. I added individual tests for every page and fixed a ton false positives. I've much happier with our coverage now.


jmervine:
Bah. Travis issue. I'll figure it out tomorrow. Might be removing the paths from the call.

jdorfman:
@jmervine want me to merge it?  Also email me about integrity.js

jdorfman:
this looks great BTW

PR #603, opened 2015-11-25
content-type checks

jmervine:
- [x] fonts
- [x] fix permission errors
- [x] add missing files
- [ ] fix content-types

### DIFF
--- a/make.js
+++ b/make.js
@@ -5,10 +5,11 @@ var path = require('path');
 var http = require('http');
 var fs   = require('fs');
 
-var MOCHA = path.join(__dirname, 'node_modules/.bin/mocha');
-var BOOTLINT = path.join(__dirname, 'node_modules/.bin/bootlint');
-var VALIDATOR = path.join(__dirname, 'node_modules/.bin/html-validator');
-var FOREVER = path.join(__dirname, 'node_modules/.bin/forever');
+var MOCHA      = path.join(__dirname, 'node_modules/.bin/mocha');
+var BOOTLINT   = path.join(__dirname, 'node_modules/.bin/bootlint');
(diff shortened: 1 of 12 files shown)


jdorfman:
I am really happy you wrote these because I am finding `.less` files that shouldn't be on the CDN e.g.:

https://maxcdn.bootstrapcdn.com/font-awesome/3.1.0/less/mixins.less

jmervine:
What about the `scss` files?

jdorfman:
`.less` & `.scss` should be ignored

jmervine:
@jdorfman before merging this, I want to clarify a few things: 

1. This will increase travis test run time to ~5 minutes.
2. This will make ~1800 requests to maxcdn.bootstrapcdn.com on each travis run (2 per pull request + 1 or 2 per merge).
3. This will add a certain level of fragility to the travis suite, as if an http connection should happen to fail (things happen on the interwebz), travis will be red.
4. Should a new file be added to the public dir on the git repo, but not the CDN, it will fail.

If you feel good with all that, go a head and merge, if not, we should discuss further.

Cheers!

jdorfman:
@jmervine would it be faster if we upgraded: https://docs.travis-ci.com/user/migrating-from-legacy/

jmervine:
@jdorfman That's not really the issue, it might make it a little faster (15-20 seconds). Last I checked though, our HTML validator fails if we use the docker based infrastructure. There is work I can do to speed this up, I'm currently synchronously iterating over all files in the directory and fetching and testing them one at a time. I could put this on a async loop that does ~5 or so at a time, but it's considerably more work / complexity / etc. Doing so would probably drop a 2 or so minutes off the total test time.

jdorfman:
@jmervine F it, I'm merging.  If we need to roll back, then we roll back.

PR #609, opened 2015-12-02
removed obsolete test

jdorfman:
merge whenever @jmervine


jmervine:
@jdorfman is there anything in `tests/jdorfman_tests` that we still need?

jdorfman:
@jmervine I want to keep the visuals and use them again in the future.  Do you think I should just remove them for now?

jmervine:
@jdorfman I, personally, wouldn't store them in the repo unless their being used.

PR #611, opened 2015-12-08
Added Boostrap version 4 alpha files

jdorfman:


jdorfman:
@jmervine should we make `text/javascript` an exception?  I see us running into this issue in the future.

jmervine:
Is `4.0.0-alpha.2` the version scheme that they're using?

jdorfman:
> 4.0.0-alpha.2 the version scheme?

@jmervine according to this: https://github.com/twbs/bootstrap/tree/v4.0.0-alpha.2 (I dropped the `v`)

PR #614, opened 2015-12-09
merging develop into master

jdorfman:
please review @jmervine


jdorfman:
@jmervine good to go?  If so just merge

PR #616, opened 2015-12-16
Adding sri to bootstrapcdn.com's link and script tags.

jmervine:
##### Outstanding Issues

- [x] Doesn't work on Chrome (tested on v47 for Mac)
- [x] Causes html-validator to file, which led to it being disabled.

@jdorfman Added as you requested. I'm getting the following errors on Chrome v47 (for Mac).

```
Subresource Integrity: The resource 'https://maxcdn.bootstrapcdn.com/font-awesome/4.5.0/css/font-awesome.min.css' has an integrity attribute, but the resource requires the request to be CORS enabled to check the integrity, and it is not. The resource has been blocked because the integrity cannot be enforced.

Subresource Integrity: The resource 'https://maxcdn.bootstrapcdn.com/bootswatch/3.3.5/united/bootstrap.min.css' has an integrity attribute, but the resource requires the request to be CORS enabled to check the integrity, and it is not. The resource has been blocked because the integrity cannot be enforced.

Subresource Integrity: The resource 'https://code.jquery.com/jquery-1.11.2.min.js' has an integrity attribute, but the resource requires the request to be CORS enabled to check the integrity, and it is not. The resource has been blocked because the integrity cannot be enforced.

Subresource Integrity: The resource 'https://maxcdn.bootstrapcdn.com/bootstrap/3.3.6/js/bootstrap.min.js' has an integrity attribute, but the resource requires the request to be CORS enabled to check the integrity, and it is not. The resource has been blocked because the integrity cannot be enforced.
```

However, this might be related to [a Chrome bug reported as of v45](https://code.google.com/p/chromium/issues/detail?id=527436). We should double check our CORS settings to be safe though.

This shouldn't be merged until this is working for sure though.

FYI: Works on Firefox.


jmervine:
Okay, @jdorfman pointed out that the tags w/ integrity were missing `crossorigin="anonymous"` (oops). So that's fixed. Still the open question on do we release without a `html-validator`. I want to checkout the source for the validator we use and see if there's a way to toggle certain checks... just haven't had time yet.

jdorfman:
@jmervine interesting.  It looks like the `html-validator` module uses the W3C Validator service.  According to the [changleog](https://validator.w3.org/whatsnew.html#v14) Subresource Integrity support has not been added causing the [error](https://travis-ci.org/MaxCDN/bootstrap-cdn/builds/97309329#L593) you pointed out.

> I want to checkout the source for the validator we use and see if there's a way to toggle certain checks... just haven't had time yet.

No, if we want to promote and advocate SRI we have to help fix the issue, not just ignore certain checks.

@jonathanKingston does the W3C and Mozilla (`html5-lint`) plan to add SRI support to their validators soon? I know it is just a "Recommendation" but 3 major browsers have implemented it.  If we want developers to adopt this, we need to make sure their tests don't break. ;)

jonathanKingston:
@jdorfman the w3cvalidator bug was logged a while back (https://github.com/validator/validator/issues/151), I will raise an issue with html5-lint as there shouldn't be any objections other than getting the work done.

Yup after two implementations it can progress... CR is pretty solid as far as most developers care, especially as this is only an additive feature of security for supporting browsers.

>  missing crossorigin="anonymous"

I'm partly responsible for that :stuck_out_tongue_closed_eyes:

jmervine:
@jdorfman (cc @jonathanKingston) I switched to `w3cjs` and built html validation in to the test suite and filtered out integrity errors.

jdorfman:
Awesome work @jmervine

PR #617, opened 2015-12-16
bootswatch 3.3.6

jmervine:
@jdorfman @thomaspark 

These are already on the cdn, just adding it to the interface now.


jdorfman:
:ship: Thanks @jmervine

PR #626, opened 2016-01-08
Syncing develop with Master

jdorfman:


jdorfman:
@jmervine can you check and merge this?  Don't want to screw anything up.  Thanks

jmervine:
Looks good. The things in scripts don't impact the running site anyway, so there's never risk in merging w/ that if you're trying to get something out.

PR #627, opened 2016-02-01
Added Bootply and Submit to integrations, updated jumbotron Tweet

jdorfman:
* Bootply (The Bootstrap Playground) is a Bootstrap Editor and Builder
* 'Submit Yours' button sends an email to @jdorfman 
* Updated the Bootstrap Twitter handle to `@getbootstrap` from `@twbootstrap`


views/_partials/header.jade
```diff
@@ -17,7 +17,7 @@
                 li: a(href="/bootlint/" + themeQs) Bootlint
                 li: a(href="/legacy/" + themeQs) Legacy
 
-            ul.nav.navbar-nav.navbar-right.visible-md.visible-lg
```
jmervine: I think these classes were designed to clean up the display on mobile. I would check this on a phone before merging to master.

jdorfman: @jmervine good call, it works on mobile.  Reason I removed that those classes is that it was hiding  `Showcase` and `Integrations`.  Is the way I did it hacky?

jmervine: Ah, probably not. :)

PR #633, opened 2016-02-08
adding http -> https redirection for production env

jmervine:
cc @jdorfman


jdorfman:
:+1:

XhmikosR:
This is nice **but** it seems to fail for the naked domain. Try https://hstspreload.appspot.com/?domain=bootstrapcdn.com which is what I aim to fix eventually with #700.

PR #634, opened 2016-02-09
Releasing sslify to production.

jmervine:


jmervine:
It would appear that the issue we were seeing before is back. I'll have to dig in before we merge this. It's tough though, because I can't reproduce locally.

jdorfman:
@jmervine k

PR #635, opened 2016-02-10
putting nav links in active mode when page is loaded

jmervine:
@jdorfman per your email

### DIFF
--- a/views/_partials/javascripts/after.jade
+++ b/views/_partials/javascripts/after.jade
@@ -48,6 +48,10 @@ script
             }
         });
 
+        (function(){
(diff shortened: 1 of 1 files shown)


jdorfman:
:+1: :ship:

XhmikosR:
What is the reasoning for this?

jmervine:
The nav button for the current page wasn't shown as active when the page was loaded. This makes it so.

XhmikosR:
We should make use of the active class then via pug.

jmervine:
I considered that when breaking things out in to their own pages, but it was actually more work/code to pass active through as a local in the route. That said, I have no strong preference one way or another.

PR #640, opened 2016-02-23
releasing SRI sha384 to production

jmervine:


jdorfman:
QA ✓

jdorfman:
@jmervine for some reason the "Quick Start" sha's are 256 & 512

image

PR #645, opened 2016-03-02
parallelize travis

jmervine:


jdorfman:
@jmervine verdict?

jmervine:
Lets let it hang and I'll look at it more later.

PR #646, opened 2016-04-11
Added Font Awesome 4.6.0, Udacity in showcase

jdorfman:


jmervine:
For future reference, I prefer to separate adding updates (e.g. fontawesome, bootstrap, etc.) from other changes, in case the PR needs to be reverted. I wouldn't take the time to do it for this though.

jmervine:
Note failing travis, perms and content-type aren't setup correctly on s3. @jdorfman

PR #657, opened 2016-04-18
Added React-Bootstrap to Integrations

jdorfman:


jmervine:
@jdorfman we may want to consider adding pagination to Integrations and Showcases if they keep growing at their current rate.

jdorfman:
@jmervine yes that and maybe randomization?

jmervine:
I wouldn't bother with Randomization, I would add new ones at the top and paginate, personally. But if you prefer randomization, I can do that as well. =D

PR #669, opened 2016-05-23
It's WordPress, dangit!

HardeepAsrani:
It should be WordPress, not Wordpress. The community is pretty serious about this. :)

Fixes: https://github.com/MaxCDN/bootstrap-cdn/issues/668


jmervine:
Good catch, thank you! Merging releasing to our staging server and once @jdorfman +1's I'll release to prod.

Cheers!

PR #670, opened 2016-05-23
Prod Release

jmervine:
- FA 4.6.3 - @jdorfman not sure why this commit isn't already on `master`, as FA 4.6.3 is showing up on www.bootstrapcdn.com?!?!?
- "It's WordPress, dangit!"

cc @jdorfman


jdorfman:
@jmervine that is weird. I'll ping you later once I have some time. @HardeepAsrani good catch!

Issue #641, opened 2016-02-23
Add referrer meta tag to website


jdorfman:
Since moving to SSL only on the site, I would like to add this meta tag.  Not high priority.

`<meta name="referrer" content="origin">`

XhmikosR:
I'd say `rel="canonical"` should be used.

https://yoast.com/rel-canonical/

jmervine:
@jdorfman Seems legit, I'll let you make the call.

jmervine:
@jdorfman Any opinions on this?

jdorfman:
@XhmikosR create a PR =)

XhmikosR:
Not sure how to do it with Jade. I still have a long to do list for this
repo....

jmervine:
Here's an example of a meta tag: https://gist.github.com/xtian/888338#file-html5boilerplate-jade-L5.

That said, I don't mind doing it if you don't have the bandwidth.

XhmikosR:
The canonical tag needs to get the current page's URL. Since you are more
familiar with Jade than me it should be easier for you.

jmervine:
@jdorfman @XhmikosR Is the suggestion `canonical` over `referrer` or both?

jmervine:
Never mind, `referrer` is already there.

XhmikosR:
My personal choice is definitely only canonical.

PR #672, opened 2016-07-20
adding canonical and removing referrer

jmervine:
RE #641 cc: @jdorfman @XhmikosR

### DIFF
--- a/app.js
+++ b/app.js
@@ -30,6 +30,8 @@ var tweets = yaml.safeLoad(fs.readFileSync(path.join(__dirname, 'config', '_twee
 app.set('port', process.env.PORT || config.port || 3000);
 app.set('views', path.join(__dirname, '/views'));
 app.set('view engine', 'jade');
+
+app.enable('trust proxy');
 app.disable('x-powered-by');
 
 // production
--- a/app.js
+++ b/app.js
@@ -31,7 +31,6 @@ app.set('port', process.env.PORT || config.port || 3000);
 app.set('views', path.join(__dirname, '/views'));
 app.set('view engine', 'jade');
 
-app.enable('trust proxy');
 app.disable('x-powered-by');
 
 // production
--- a/routes/index.js
+++ b/routes/index.js
@@ -1,7 +1,12 @@
 'use strict';
 
 function appendLocals(req, res) {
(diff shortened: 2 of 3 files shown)


XhmikosR:
I can't check this thoroughly but the concept seems goods to me. As long as
the variables are properly set that is.

On a side note we should also look into adding social tags later.

jmervine:
@XhmikosR You can see the result here: https://bootstrapcdn-dev-pr-672.herokuapp.com/

For example https://bootstrapcdn-dev-pr-672.herokuapp.com/bootlint/ has:

```
<link rel="canonical" href="http://bootstrapcdn-dev-pr-672.herokuapp.com/bootlint/">
```

XhmikosR:
Why not https?

jmervine:
@XhmikosR That's actually a good question, it should be. Perhaps the Protocol isn't being passed on correctly through Heroku's routing layer? I'll need to investigate.

jmervine:
I think I found the answer here, testing it now... http://stackoverflow.com/a/16405622/832409

XhmikosR:
@jdorfman please look into adding me to BootstraCDN. We can always chat
with Justin via Hangouts.

jmervine:
I still want to dig in to why this is occurring, but for now, I'm hard coding it to `https` as we force `https` in production anyway.

jmervine:
Better...
```
<link rel="canonical" href="https://bootstrapcdn-dev-pr-672.herokuapp.com/bootlint/">
```

XhmikosR:
It's I need strange since the protocol is https.

Might be better investigating this  before merging.

jmervine:
Fair. I'll have to table it then. I'm fairly certain it's something behind the scenes with the Heroku deploy. It shouldn't be too hard to figure out.

jmervine:
Yeah, so [that](https://github.com/MaxCDN/bootstrap-cdn/pull/672/commits/713bc46a307a1fdc22c9a43f8008f855a82f61ff) seems to have done the trick. Looks like Express isn't setting the header correctly on forward. Based on what I'm reading `enable('trust proxy')` should have worked.

jdorfman:
🚢 'd

PR #677, opened 2016-07-25
Added Bootstrap 3.3.7

jdorfman:


jmervine:
@jdorfman This was failing b/c you need to upload files to both "bootstrap" and "twitter-bootstrap" for the current suite to pass. I uploaded the files and merged. I'll make sure everything passes before moving to prod.

PR #681, opened 2016-07-27
Use sha384 for jQuery's hash.

XhmikosR:
@jmervine: can you do the same for /stylesheets/style.css and push to the merge and then merge it?


jmervine:
@XhmikosR Yeah, I'll update everything, everywhere to use 384, over both 256 and 512.

XhmikosR:
Only stylesheets/style.css is left.

jmervine:
Ah, I see, for some reason I though the assets in public were still 256/512.

PR #682, opened 2016-07-27
Require node.js >= 4.0.

XhmikosR:


jmervine:
The issue with this is I wanted to match the node version that it will be running on in Heroku. I'm not oppose to this change, but can we also include `5` when running on Travis.

XhmikosR:
So Heroku runs 5? Can we make it use 4?

jmervine:
Probably, I'll look it.

jmervine:
Yes, here... https://devcenter.heroku.com/articles/nodejs-support#specifying-a-node-js-version

I can make the update, do you have a specific version of 4.x you want to use?

XhmikosR:
I'd say the latest LTS, 4.4.7 at the moment. Can't this process be automated to keep using the latest LTS at the time?

jmervine:
From that doc...

> You can also specify a semver range (4.x). This is not recommended in production, where runtime versions should not float.

jmervine:
That said, I'm cool with using `4.x`. We'd have to update when the next LTS major version is bumped.

XhmikosR:
I think that is the whole idea of LTS.

jmervine:
@XhmikosR made the change

jmervine:
worked 

```
jmervine@laptop:~$ heroku run -a bootstrapcdn-dev node --version
Running node --version on ⬢ bootstrapcdn-dev... up, run.8560
v4.4.7
```

PR #683, opened 2016-07-27
Lint tweaks.

XhmikosR:
We should eventually add ESLint.

`mktemp` is unused so I removed it.


XhmikosR:
`mktemp` is unused so I removed it.

PR #687, opened 2016-07-27
Production Release

jmervine:
See commits.


jmervine:
I want to include #688 before releasing this.

PR #691, opened 2016-07-28
Fix Bootlint errors

XhmikosR:
@jmervine: as you can see, this wasn't caught by our Bootlint test. We need to run Bootlint for **all** pages.


jmervine:
@XhmikosR Totally agree. Linting was setup when it was a single page app and hasn't been updated. Creating a card for it.

PR #701, opened 2016-07-28
Remove class and move it to style.css.

XhmikosR:
@jmervine: we need to update the SRI hash for style.css. How do you generate them? With integrity.js? Because I'm getting tons of changes in hashes locally.

### DIFF
--- a/config/_config.yml
+++ b/config/_config.yml
@@ -13,7 +13,7 @@ favicon:
 stylesheets:
   style:
     uri: /stylesheets/style.css
-    sri: sha384-Hp4W0rZuGv3wWe7zJ2e1G/1/6sfDQI44YJ6lKrz+atPau51Jhz5UOGlFzYNZ0NqU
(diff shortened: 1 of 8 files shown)


XhmikosR:
Oh, also, maybe we should automate our assets SRI hash generation to save us some time in the future.

jmervine:
@XhmikosR `scripts/integrity.js` does that for all assets (Bootstrap, etc.) and updates config.json,  it wouldn't be that hard to expand it. We should also probably add a make task for it. Adding an issue to do so.

XhmikosR:
Weird, because I get tons of changes when I run it. Can you run it and compare the changes?

jmervine:
In the mean time, can you run `scripts/sri.js` on this file and update the config?

XhmikosR:
Wait for CI to finish, and check the deployment just to be safe.

I still believe we should check the hash stuff because I see a lot of differences even when I change everything to use LF.

jmervine:
Fair, most of the SRI's are being tested, I believe, but perhaps not style.css, that said, it should change with an change to the file.

XhmikosR:
I think we are all set.

PR #702, opened 2016-07-28
Use `https` everywhere.

XhmikosR:
BTW, I hope the Twitter link changes are OK, i.e. don't break any metrics you might have.

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,7 +13,7 @@
 
 ## Requirements
 
-1. [Node.js](http://nodejs.org/)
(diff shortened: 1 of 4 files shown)


jmervine:
@jdorfman Are we actually using the GA on this? If so, this will probably break it -- starting a new entry going forward, but it's not a terrible idea to just get it done.

XhmikosR:
How about I just skip the Twitter file completely?

jdorfman:
@XhmikosR @jmervine as long as it doesn't break GA we good

PR #706, opened 2016-07-28
Switch to pugjs

XhmikosR:
Issues:

- [x] HTML is no longer minified
- [x] uglify-js filter doesn't seem to work anymore

For uglify, we could move everything to an external JS, with cache headers, cache busting and so on.


XhmikosR:
OK, for the first issue, it's something wrong in make.js. Maybe `NODE_ENV` needs to be set for `make start`?

jmervine:
I typically use `npm run run` locally, for `make start`, it just runs `forever` and the `NODE_ENV` defaults to development here: https://github.com/MaxCDN/bootstrap-cdn/blob/develop/app.js#L10. It should be okay.

jmervine:
Oh, yeah, minification will only occur if `NODE_ENV=production`, that's intentional for dev and debugging purposes. If you view source @ https://www.bootstrapcdn.com, you'll see that it's working as expected.

XhmikosR:
Yeah  so we need to set the var to production for make start IMO. In order
to emulate the production environment as close as possible.

Now regarding this PR. Tests only fail but if you check the HTML code the
javascript is minified. Try it by reverting the relevant patch locally.

It seems they changed the interpolation, but I can't figure out the
solution yet. A workaround would be to use hardcoded values in GA.

jmervine:
That's your call, I don't use make start, it was only created for production mode when we were on Digital Ocean, it's not used else where.

XhmikosR:
I still can't make the HTML output minified locally for the `make start` target even after setting up the env variable to production.

Can you suggest a patch for this?

As for this PR, I just disabled uglify for GA only and seems to be OK.

jmervine:
The app won't run locally with NODE_ENV set to production because https redirection is turned on [here](https://github.com/MaxCDN/bootstrap-cdn/blob/develop/app.js#L39). To enable minification locally, comment out [this line](https://github.com/MaxCDN/bootstrap-cdn/blob/develop/app.js#L41).

To be able to toggle this, we'd have to put in a secondary environment variable that either disables https redirection or enables minification independently of NODE_ENV. This method, while a little hacky, is fairly common practice.

To do this, I'd do something like...

```node
// production
if (env === 'production' && process.env.DISABLE_HTTPS === undefined) {
    app.use(enforce.HTTPS({ trustProtoHeader: true }));
}

if (env === 'production') {
    app.use(logger('combined'));
} else {
    app.locals.pretty = true;
    app.use(logger('dev'));
    app.use(errorHandler({ dumpExceptions: true, showStack: true }));
}
```

You should then be able to start the app like so...

```bash
$ NODE_ENV=production DISABLE_HTTPS=true make start
```

Hope that helps.

jmervine:
On second though, let's just set this up so I can force `pretty` when I need it, let so...

```node
// production
if (env === 'production') {
    app.use(logger('combined'));
    app.use(enforce.HTTPS({ trustProtoHeader: true }));
} else {
    app.use(logger('dev'));
    app.use(errorHandler({ dumpExceptions: true, showStack: true }));
}

if (process.env.PRETTY_HTML === 'true') {
    app.locals.pretty = true;
}
```

I can update this PR or submit a new one if you'd like. LMK

XhmikosR:
Definitely a new PR :)

But, IMO we should have the `start` target be as close to the production one. I see no point in configuring the minify option when you can run the `run` target...

XhmikosR:
Yes, but locally, I don't have the variable set. And I want to automate this so that we are able to check things locally too.

jmervine:
http://stackoverflow.com/questions/9249830/how-can-i-set-node-env-production-in-windows

It looks like you can set them.

XhmikosR:
Yes, I'm aware of that. I don't want to do it manually.

jmervine:
Okay, then I'll flip it so `development` needs to be set manually. PR incoming.

jmervine:
👍 🚢 So happy to get this in, that ❌ has been bugging me. =D

PR #709, opened 2016-07-29
Update .travis.yml.

XhmikosR:
* specify git depth 10
* cache node_modules folder

Should save us around 30 seconds per build.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,10 +1,21 @@
+git:
+  depth: 10
+
 language: node_js
+
 node_js:
   - "4"
-script: make travis
+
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -1,10 +1,21 @@
+git:
+  depth: 10
+
 language: node_js
+
 node_js:
   - "4"
-script: make travis
+
+script: node make travis
```
jmervine: Because of how the `Makefile` setup, these are identical. So yeah, I'm good either way. :D

jmervine:
👍 🚢

PR #712, opened 2016-07-29
Clean up obsolete files and README.md.

jmervine:


XhmikosR:
Maybe delete tests/stubs too.

XhmikosR:
And squash :P

README.md
```diff
@@ -76,27 +53,6 @@ node make status
 tail -f ./logs/server.log
```
XhmikosR: Is this still valid?

XhmikosR: If not, .gitignore has some old entries we should remove.

jmervine: It's not, but that's actually a bug. When I setup the task in `make.js` I forgot to have it log locally. It currently logs to forevers default location which is accessible via `node ./node_modules/.bin/forever logs app.js`.

jmervine:
You can now squash while merging, which I kind of like...

screen shot 2016-07-29 at 10 52 45 am

XhmikosR:
I know but if I merge it and squash, will it keep you as the author?

jmervine:
I believe so, I've used this feature a bunch in other apps and if it does mess with the author, I didn't notice it.

XhmikosR:
Let me try it then.

PR #713, opened 2016-07-29
Remove bootlint query support.

XhmikosR:
It doesn't work and a proper Bootlint run should be made against all pages.

/CC @jmervine


jmervine:
👍

PR #717, opened 2016-07-30
Normalize line endings.

XhmikosR:
So, public/twitter-bootstrap/2.1.1/js/bootstrap.min.js seems to have been committed with CRLF or mixed line endings.

/CC @jmervine


jmervine:
Updated s3 and purged cache.

XhmikosR:
So everything passes now?

I'd like to see some numbers for this specific file although if doubt
people were using SRI so they won't end up with broken scripts.

jmervine:
We actually should release this ASAP just in case someone was using it w/ SRI (which I also doubt).

Issue #705, opened 2016-07-28
Ensure integrity.js includes all assets


jmervine:
- Ensure integrity.js includes all assets
- Add make task to run it

jdorfman:
Reopen if needed

PR #718, opened 2016-07-30
Add cache busting for our files

### LINKED ISSUE #716: Add cache busting for our files

jmervine:
Fixes #716

### DIFF
--- a/lib/helpers.js
+++ b/lib/helpers.js
@@ -1,7 +1,10 @@
 'use strict';
 
+// create datestamp on startup
+var cacheBuster = Date.now().toString();
+
 var sri = require('sri-toolbox');
-var fs  = require('fs');
+var fs = require('fs');
 
 function sriDigest(file, fromString) {
     file = fromString ? file : fs.readFileSync(file);
@@ -40,14 +43,39 @@ function randomTweets(tweets, limit) {
     return retTweets;
 }
 
+function cssWithBuster(uri, sri) {
+    return '<link rel="stylesheet" href="' + uri + '?' + cacheBuster +
+        '" integrity="' + sri + '" crossorigin="anonymous">';
+};
+
+function jsWithBuster(uri, sri) {
+    return '<script src="' + uri + '?' + cachBuster +
+        '" integrity="' + sri + '" crissorigin="anonymous"';
+};
+
+function imgWithBuster(uri, attributes) {
+    var attrs = '';
+
+    for (var key in attributes) {
+        attrs += ' ' + key + '="' + attributes[key] + '"';
+    }
+
+    return '<img src="' + uri + '?' + cacheBuster + '"' + attrs + '>';
+};
+
 module.exports = {
     theme: {
         selected: selectedTheme,
-        fetch:    getTheme
+        fetch: getTheme
     },
     randomTweets: randomTweets,
     sri: {
         digest: sriDigest
+    },
+    cacheBuster: {
+        css: cssWithBuster,
+        js: jsWithBuster,
+        img: imgWithBuster
     }
 };
 
--- a/lib/helpers.js
+++ b/lib/helpers.js
@@ -6,6 +6,10 @@ var cacheBuster = Date.now().toString();
 var sri = require('sri-toolbox');
 var fs = require('fs');
 
+function buster(uri, sri) {
+    return uri + '?' + (sri ? sri : cacheBuster);
+}
+
 function sriDigest(file, fromString) {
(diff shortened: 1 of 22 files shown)


lib/helpers.js
```diff
@@ -1,7 +1,10 @@
 'use strict';
 
-var sri = require('sri-toolbox');
-var fs  = require('fs');
+// create datestamp on startup
+const cacheBuster = Date.now().toString();
```
XhmikosR: You should probably revert these to `var` since we don't use ES6 elsewhere.

XhmikosR:
I don't like the query approach. I prefer the `filename.hash.ext` approach personally.

views/_partials/integrations.pug
```diff
@@ -1,9 +1,11 @@
 hr
 br
 a(href=item.url, target="_blank")
-    img.img-responsive.thumbnail.center-block(src=item.img, alt=item.name)
+    != helpers.cacheBuster.img(item.img, { alt: item.name, class: "img-responsive thumbnail center-block" })
 h3
     a(href=item.url, target="_blank")=item.name
-| Platform:
```
XhmikosR: The space is needed here.

views/_partials/showcase.pug
```diff
@@ -1,10 +1,10 @@
 hr
 br
 a(href=item.url, target="_blank")
-    img.img-responsive.thumbnail.center-block(src=item.img, alt=item.name)
+    != helpers.cacheBuster.img(item.img, { alt: item.name, class: "img-responsive thumbnail center-block" })
 h3
     a(href=item.url, target="_blank")=item.name
-| Libraries:
```
XhmikosR: And here.

lib/helpers.js
```diff
@@ -40,14 +43,39 @@ function randomTweets(tweets, limit) {
...
+
+function jsWithBuster(uri, sri) {
+    return '<script src="' + uri + '?' + cachBuster +
+        '" integrity="' + sri + '" crissorigin="anonymous"';
+};
+
+function imgWithBuster(uri, attributes) {
+    var attrs = '';
+
+    for(var key in attributes) {
```
XhmikosR: Missing space after `for`.

lib/helpers.js
```diff
@@ -40,14 +43,39 @@ function randomTweets(tweets, limit) {
...
+        attrs += ' ' + key + '="' + attributes[key] + '"';
+    }
+
+    return '<img src="' + uri + '?' + cacheBuster + '"' + attrs + ' />';
```
XhmikosR: No need to use `/>`. It's already self closing in HTML5, The same applies for `link`.

views/_partials/integrations.pug
jmervine: Right! Good catch, I have my EOL space removal set to ignore jade, need to update it to ignore pug too.

lib/helpers.js
jmervine: Okay, sure.

XhmikosR:
On the other hand, my suggestion depends on #705, so I guess the query parameter solution is as good as it can get now.

jmervine:
@XhmikosR I like that as well, and should we move to a precompiled static site setup, fronted by a CDN, I'd consider that. I think in the mean time though, this is a functional and acceptable solution. 

What I don't like is that it resets the cache buster on restart. I could write something in the startup that generates a sha of each file. I also considered using the sri for the css file.

jmervine:
Addressed as suggested and rebased with `develop` so it should be right now.

XhmikosR:
Please skip the Bootstrap file. I'll see later if it's something specific to nix.

On a side note, the tests fail.

XhmikosR:
I just tested on Debian 8.5 and git 2.1.4 a fresh clone and works fine, i.e no modified files.

jmervine:
I'm trying to skip it, I haven't been able to figure out how, even with the doc you sent. Do you want to checkout this branch and take a stab at it?

XhmikosR:
I can try rebasing it skipping the offending file. Just note that you will
need to reset your branch after fetch.

XhmikosR:
Done. Only thing is if we'll keep the whole tag in the JS or not.

jmervine:
I'm not seeing your commit.

XhmikosR:
There is no commit to see :p I just rebased against develop and skipped the
offending patch.

jmervine:
RE the tag in the JS, I debated that. Since you mention it, it's probably better to just output the url with buster, then I don't need one function for each either. I'll make the change.

PR #727, opened 2016-08-04
Tweak scripts/integrity.js

XhmikosR:
* specify `lineWidth: 110`. The default is 80, so in order not to have random changes, increase it to 110.
* take into consideration the special case for bootstrap folder

/CC @jmervine: should we go with `80` for `lineWidth`?

### DIFF
--- a/scripts/integrity.js
+++ b/scripts/integrity.js
@@ -16,8 +16,9 @@ fs.createReadStream(configFile)
     .pipe(fs.createWriteStream(configFile + '.bak'));
 
 function buildPath(d) {
-    return path.join(basedir, 'public',
-        d.replace('https://maxcdn.bootstrapcdn.com/', ''));
+    d = d.replace('/bootstrap/', '/twitter-bootstrap/')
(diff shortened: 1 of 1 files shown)


scripts/integrity.js
```diff
@@ -75,4 +77,4 @@ for (var i = 0; i < config.fontawesome.length; i++) {
     }
 }
 
-fs.writeFileSync(configFile, yaml.dump(config));
+fs.writeFileSync(configFile, yaml.dump(config, { lineWidth: 110 }));
```
jmervine: What's this do?

XhmikosR: Please check the whole commit for this patch.

jmervine: I read over everything else and it looks good. I just wasn't familiar with the `lineWidth` option. I'll google. :P

XhmikosR: Just a JS yaml option which is why I ask you if we should stick with the
default because that is the cause for the changes in config.

jmervine:
👍 🚢

PR #728, opened 2016-08-04
Clean up package.json dependencies.

XhmikosR:
Also moved a few to devDependencies since they seem to be used in tests only.

/CC @jmervine


jmervine:
Hopefully rebasing this should fix it.

jmervine:
👍 🚢

PR #733, opened 2016-08-06
moving inline Javascript to separate file

### LINKED ISSUE #365: Consider moving inline Javascript to separate file and minify it
We could use something like this (untested)  http://stackoverflow.com/a/16325626

jmervine:
Fixes #365

- Moving inline js to separate file
- Added minify middleware for express to automatically minify out going css and js

@XhmikosR


XhmikosR:
Needs some work. Needs SRI for main.js plus I'd move GA there too.

package.json
```diff
@@ -19,9 +19,13 @@
     "compression": "^1.5.2",
     "errorhandler": "^1.4.3",
     "express": "^4.13.3",
+    "express-minify": "^0.2.0",
     "express-sslify": "^1.0.1",
-    "jstransformer-uglify-js": "^1.1.1",
+    "fs-walk": "0.0.1",
```
XhmikosR: These seem unrelated changes.

package.json
```diff
@@ -19,9 +19,13 @@
...
+    "html-validator": "^1.0.0",
+    "htmlencode": "0.0.4",
     "js-yaml": "^3.4.2",
+    "jstransformer-uglify-js": "^1.1.1",
```
XhmikosR: And this shouldn't be needed anymore, right?

jmervine: `express-minify` may use it. I'll verify.

package.json
jmervine: I accidentally branched off master, then cherry picked my commits to a branch off develop, must have missed that when cleaning up. I'll look at it sometime over the weekend.

package.json
jmervine: Actually it was already there just higher up, but probably not. As I said, I'll verify.

XhmikosR: It was there sure because we used it before these changes :P

XhmikosR:
Rebased and cleaned up the branch.

We can't use express-minify since the SRI hashes will differ due to minification.

After you review this, please ping me to make one last rebase.

jmervine:
main.js has SRI. What am I missing here?

For GA, I don't want to move it in because I want the config to stay abstracted  it needs to be in the template for interpolation. I guess we could assign the vales to js values in the template prior, but that seemed overkill.

jmervine:
I just saw your last comment ... On a phone and missed it ... That totally makes sense. 

Perhaps generating the SRI from the output from express-minify?

jmervine:
I realize now, I may not have pushed my last commit w/ SRI for main.js. Either way, I Think this looks good.

I'm going to work on "on the fly" generating SRI for main.js and style.css on a branch off of this branch. If it looks good, we can merge in to this.

jmervine:
@XhmikosR Here's what I'm thinking for super simple sri generation on the fly for `style.css` and `main.js`. It's cached in memory (which I verified) and pretty quick even without the caching.

https://github.com/MaxCDN/bootstrap-cdn/compare/jmervine-external-js...jmervine-external-js-sri

Other thoughts...

1. I don't think we want to keep the `crossorigin="anonymous"` for these two files, as they should only be served off `www.bootstrapcdn.com`.
2. We might actually want etags for these, as they're not served off the CDN.
3. For GA, as I mentioned, I think we should leave that where it is for now. I really want to keep that configured and flag-able in `config/_config.yml`.

XhmikosR:
I'll comment on that branch later looks pretty good but I want to verify

Cross origin is needed by the specs.

For GA I think we can just use hard-coded values and even remove the
related tests thus removing the entries from config completely.

I don't see why we would ever need ETags if we set cache properly...

PR #734, opened 2016-08-06
removing etag and adjusting cache-control headers

### LINKED ISSUE #699: Remove ETags
```js
app.set('etag', false);
```

The CDN files themselves have this header too.

### LINKED ISSUE #708: Revisit cache control header
For HTML we shouldn't have `2592000` s = 30 days. Maybe `3600` (1 hour) tops.

Now, for the static assets, CSS, images, JS, we **should** use a long time. But in order to prevent issues, we should add cache busting.

jmervine:
Fixes #708, fixes #699 

Had to reorder mw a little to ensure that things were applied in the proper order. Additionally, I removed the `Expires` header, as without `ETag`, it shouldn't be necessary any more.

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,13 +63,12 @@ Key Overview:
 2. `theme`: Integer value of the array index from the `bootswatch` section below.
 3. `authors`: Array of Author Strings
 4. `description`: String containing the meta description of the site.
-5. `favicon`: Hash containing the favicon path.
-6. `google_analytics`: Hash containing GA `account_id` and `domain_name`.
-7. `stylesheets`: Array containing stylesheet files to be loaded at the top of the site.
-8. `javascripts`: Array containing javascript files to be loaded either `before` (at the top) or `after` (at the bottom) of the site.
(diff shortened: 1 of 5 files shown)


XhmikosR:
It seems the static files still have ETag set.

XhmikosR:
Also, I do believe we should set `Expires` too along with `Last-Modified` (`lastModified` expressjs option) for HTML.

XhmikosR:
I added a couple more patches @jmervine. Ideally I guess I should make a new PR, but you know :p

XhmikosR:
So from my suggestions above, only `Last-Modified` is left for HTML to add.

XhmikosR:
I wonder if we could just use https://github.com/h5bp/server-configs-node.

jmervine:
I left the Etag on the static assets because they're not served through the cdn. Not while we're using s3.

jmervine:
These all look good. I think Travis is failing because of bootlint minify changes. Merge that and rebase and it should be good.

PR #736, opened 2016-08-07
Add ESLint support

### LINKED ISSUE #684: Add ESLint
I have a sort of working config, we need to discuss some style options.

I'll post the details tomorrow.

XhmikosR:
First pass, **do not merge** until ready.

Non-whitespace diff: https://github.com/MaxCDN/bootstrap-cdn/pull/736/files?w=1

/CC @jmervine for thoughts on the rules, scripts etc. Note that I quickly added this in npm scripts, should probably be in make.js.

Fixes #684.


scripts/integrity.js
```diff
@@ -30,51 +31,60 @@ function exists(file) {
...
+        }
     }
-}
+})();
 
 // bootlint
-for (var i = 0; i < config.bootlint.length; i++) {
-    var bootlint = config.bootlint[i];
-    var file     = buildPath(bootlint.javascript);
+(function() {
```
jmervine: Why the closures? I ask, not to be contrary, but more for my own education. I never know when they're appropriate and when they're not.

tests/functional_test.js
```diff
@@ -1,41 +1,41 @@
...
+var digest = require(path.join(__dirname, '..', 'lib', 'helpers')).sri.digest;
 var helpers = require(path.join(__dirname, 'test_helper'));
-var config  = helpers.config();
+var config = helpers.config();
 
 var expectedHeaders = {
-    date: undefined,
-    etag: undefined,
-    expires: undefined,
+    'date': undefined,
```
jmervine: Why quote these?

scripts/integrity.js
XhmikosR: I went this way because `i` and `file` variables were being redefined in the whole file. So, with IIFE we don't pollute the global scope.

tests/functional_test.js
XhmikosR:
```
"quote-props": [2, "consistent"]
```

Feel free to check out and suggest any other options. I'm not tied to these ones, just tried to be as strict as possible and close to my own personal style.

jmervine: I see, so if some are quoted, it recommends that all in the object are for consistency. That makes sense. I have no strong feelings on it, was just wondering.

scripts/integrity.js
jmervine: Makes sense. Thanks.

tests/functional_test.js
XhmikosR: http://eslint.org/docs/rules/quote-props

It shouldn't matter in our case that is why I went with consistent.

jmervine:
This all looks great. My only comment is that I really prefer the tabularized alignment where it makes sense. But I'm willing to forgo it in favor of consistency.

XhmikosR:
I tried to keep that but couldn't find the right config.

Also, there are a few more warnings and errors that we need to sort.

This branch will take some time but should help keep the JS codebase consistent :)

XhmikosR:
OK, I think it's `"exceptions": { "VariableDeclarator": true } `. I will try it. In the meantime, check out the branch (I won't rebase until we are ready to merge it) and see if you can sort the rest.

XhmikosR:
@jmervine: I'll need to rebase this later and revert the alignment changes. If you have something else to commit here let me kmow.

jmervine:
Nothing at this time. :D

XhmikosR:
I'm gonna finish this tomorrow...

public/assets/js/main.js
```diff
@@ -1,5 +1,6 @@
 /* eslint-env browser */
 /* exported tryIt, toggleCode */
+'use strict';
```
XhmikosR: This isn't right BTW, I will take care of it tomorrow with stricter options too.

XhmikosR:
@jmervine: I added a few more rules.

About main.js, we need to refactor it to use IIFE and be explicit with global variables. Also, check the new errors when you have some time along with the new rules that I have added and are set to `0`.

XhmikosR:
Rebased and pushed. Should be ready for now. We can tweak it later.

jmervine:
I'll merge as soon as CI finishes.

PR #739, opened 2016-08-07
Add .editorconfig file.

XhmikosR:


jmervine:
If this helps you, merge it for sure. :D

jmervine:
Note on this for anyone that uses vim and happens to come across it... see: https://github.com/editorconfig/editorconfig-vim

PR #745, opened 2016-08-12
Add express-minify.

### LINKED ISSUE #744: Minify our assets

XhmikosR:
Basically reverts commit ab439e7d88d444dae7d3b8447eb2d21bf3a09d25.

Doesn't seem to work though, the SRI hashes don't match.

Fixes #744.

### DIFF
--- a/app.js
+++ b/app.js
@@ -2,23 +2,24 @@
 
 var env = process.env.NODE_ENV || 'development';
 
-var path        = require('path');
-var fs          = require('fs');
-var http        = require('http');
-var express     = require('express');
-var yaml        = require('js-yaml');
-var compression = require('compression');
-var sitemap     = require('express-sitemap');
+var path         = require('path');
+var fs           = require('fs');
+var http         = require('http');
+var express      = require('express');
+var yaml         = require('js-yaml');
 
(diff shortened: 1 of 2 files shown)


jmervine:
Because the files are run through [`generateSRI`](https://github.com/MaxCDN/bootstrap-cdn/blob/develop/routes/index.js#L24-L30), the SRI should be right...

XhmikosR:
Must be an issue with the order. I can't think of anything else.

app.js
```diff
@@ -54,6 +55,7 @@ if (env === 'production') {
 
 // middleware
 app.use(compression());
+app.use(minify());
```
jmervine: I believe this should be above `compression()`.

XhmikosR: I just followed their docs https://github.com/SummerWish/express-minify#basic-usage

jmervine: I see, it was just a guess.

jmervine:
@XhmikosR Did the revert fix the SRI issue?

XhmikosR:
Nope this is still broken.

On a side note, it seems we don't use any compression for static assets like CSS and JS files.

PR #747, opened 2016-08-12
Add CSP header.

### LINKED ISSUE #746: Add CSP header

XhmikosR:
Needs some work since ads are being blocked currently. Also, it could be simplified using wildcards maybe.

@ScottHelme: Can I have more than one projects in report-uri.io? I'd like to add this as report only for starters but I already have another domain set in my settings.

Fixes #746.

### DIFF
--- a/app.js
+++ b/app.js
@@ -17,6 +17,7 @@ var logger       = require('morgan');
 var serveStatic  = require('serve-static');
 var errorHandler = require('errorhandler');
 var enforce      = require('express-sslify');
+var csp          = require('express-csp');
 
 var helpers      = require('./lib/helpers');
 var routes       = require('./routes');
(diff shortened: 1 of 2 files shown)


XhmikosR:
Rebased and seems to work fine now.

I'm not sure if it can be simplified. Maybe @ScottHelme can share his thoughts.

ScottHelme:
You can collect reports for multiple site on https://report-uri.io if you need. In the "Hosts to collect reports for" field on the Filters page you can specify multiple domains. For example: 

    *.example.com example.com *.example.co.uk example.co.uk

You can then filter on these different values in the Reprots page using the drop down options.

<br/>
As for the CSP, I'd strongly advise whitelisting inline script with a nonce instead of unsafe-inline. You can leave the unsafe-inline there for backwards compatibility with CSP 1 compliant browsers, but in CSP 2/3 compliant browsers the nonce value would override unsafe-inline and offer much better protection.

XhmikosR:
The problem with nonce is that it requires more changes from our side...

Perhaps we should just go with the current PR and add an issue to switch to nonce later.

jmervine:
@XhmikosR I'll defer to you on this. I don't have a strong opinion. =D

XhmikosR:
OK, so here's the situation.

1) we have a lot of inline JS code. I moved some but there's more
2) unfortunately radar.js is using `eval` and we need to specify `unsafe-eval`
3) for `style-src` I couldn't make Google Fonts work with `nonce`; only with `unsafe-inline` it works for me

@ScottHelme: if you have ideas how to proceed, please let me know. Removing `unsafe-inline` for JS can be done with some work, but see 2 and 3 from my above points.

ScottHelme:
@XhmikosR can you show me how you're loading the Google Font?

As for radar.js, bug it with them, but there's not much you can do sadly.

XhmikosR:
@ScottHelme: Google Fonts are loaded with an `@import` rule due to the way Bootswatch ships the themes. See this for example  https://github.com/MaxCDN/bootstrap-cdn/blob/develop/public/bootswatch/3.3.7/united/bootstrap.css#L1

For radar.js @jdorfman please bug them to change their code and stop using `eval`.

XhmikosR:
Hmm, might be something else that's trying to change the styles and not Google Fonts.

ScottHelme:
Yeah the `@import` rule should just be a domain whitelist entry. Let me know if you track it down.

XhmikosR:
It seems it's due to `script(async src='https://platform.twitter.com/widgets.js')`. If I remove it I get no error with nonce for style-src.

XhmikosR:
So, the only inline JS now is Google Analytics which we can use a nonce with express-csp and the tracking code the jumbotron links have.

Does anybody know how we can move those to JS? I have never used this before.

views/bootswatch.pug
```diff
@@ -13,15 +13,15 @@ block content
                         - if (i === helpers.theme.selected(config, theme))
                             button.btn.btn-primary(type="button", disabled) Active!
                         - else
-                            button.btn.btn-primary(type="button", onclick="tryIt(" + i + ")") Try it!
+                            button.btn.btn-primary(type="button", data-theme=i) Try it!
```
XhmikosR: @jmervine: I'm thinking, we should be able to do this without JS via Pug. Like doing `a href=/?theme=x`.

jmervine: My thought in doing it though client js was supporting static site generation down the road. However, at this point, I think I would prefer to put the entire site behind a CDN, which would work nicely with the method you're suggesting.

I'll leave it up to you, if you want to play with it on the short term, otherwise, I might be able to look at tweaking it in a few weeks.

XhmikosR:
So, does anybody have any idea how to target those GA events without using an `onclick` event in HTML?

Also, how about we move the GA snippet to the JS file?

XhmikosR:
So @jmervine what do you think? I move GA to main.js too.

jmervine:
👍

XhmikosR:
@jmervine: I need to find a way for the GA tracking stuff we have in jumbotron to move it to JS. Do you have any hint for that?

jmervine:
You're talking about the `onclick` events, yeah? If so, we could perhaps use JS to create those attributes on load.

XhmikosR:
Please check the other PR. I attempted to do this but I have no idea if it
works.

jmervine:
I saw it, I'd need to do some testing to see. I'd probably add an `alert` for testing just to make sure it's being run properly. Then we'll need to coordinate with @jdorfman and make sure the data is still coming through after release.

PR #749, opened 2016-08-12
Fix navbar active items when using themes.

XhmikosR:


XhmikosR:
@jmervine: I added a second patch sort of unrelated to this issue but I find this cleaner. If you have any reason to use `/bootstrap/` hardcoded let me know.

jmervine:
:+1:

PR #751, opened 2016-08-13
Move .active nav menu class to pug.

### LINKED ISSUE #742: Move `.active` nav class to pug
Here is how I'm thinking of doing it:

1. Have an object with the title/url of the pages
2. Check if the current menu item's title/url matches, then add .active otherwise no class.

/CC @jmervine

XhmikosR:
Fixes #742.

### DIFF
--- a/public/assets/js/main.js
+++ b/public/assets/js/main.js
@@ -34,19 +34,6 @@
             });
         }
     });
-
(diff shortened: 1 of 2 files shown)


jmervine:
👍

PR #757, opened 2016-08-17
disabling cache buster in non-prod environments

### LINKED ISSUE #741: Disable caching for development

jmervine:
@XhmikosR Fixes #741

### DIFF
--- a/lib/helpers.js
+++ b/lib/helpers.js
@@ -7,7 +7,7 @@ var sri = require('sri-toolbox');
 var cacheBuster = Date.now().toString();
 
 function buster(uri) {
-    return uri + '?' + cacheBuster;
+    return uri + (process.env.NODE_ENV === 'production' ? '?' + cacheBuster : '');
 }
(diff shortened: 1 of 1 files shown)


XhmikosR:
I don't like this solution. It's one thing disabling caching and another remove the cache busting.

jmervine:
What's the difference?

XhmikosR:
Because you are not disabling caching that way...

jmervine:
I don't understand the issue here. `shift-` refresh clears the browser cache, and really cache busters shouldn't matter locally anyway, as it's not being served through the CDN.

Development needs to be a second class citizen. Minor changes, okay, but making larger changes -- e.g. headers, etc. -- for development, I don't agree with.

I'm going to close this, because as you say, it really doesn't help in any way.

XhmikosR:
I agree. It's just at the moment, development is having issues and I'm trying to pinpoint the cause.

PR #759, opened 2016-08-22
Remove Makefile info.

XhmikosR:


XhmikosR:
@jmervine: I think we should also drop `npm run` tasks. It's just simpler to only support one way.

jmervine:
👍 I'm probably not going to have time for that this week, but if you want to take a crack at it, feel free. Otherwise, I'll look at it when I can.

PR #760, opened 2016-08-26
Add AppVeyor

XhmikosR:
Since, I'm a Windows user and recently snyk messed up everything, I'd like to have AppVeyor for Windows CI.

Also, it shouldn't hurt having 2 CIs running.

/CC @jdorfman @jmervine

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,6 +2,7 @@
 
 [![Follow BootstrapCDN on Twitter](https://img.shields.io/badge/twitter-@getBootstrapCDN-55acee.svg?style=flat-square)](https://twitter.com/getbootstrapcdn)
 [![Linux Build Status](https://img.shields.io/travis/MaxCDN/bootstrap-cdn/develop.svg?label=Linux%20build&style=flat-square)](https://travis-ci.org/MaxCDN/bootstrap-cdn)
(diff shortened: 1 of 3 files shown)


XhmikosR:
It seems for some reason, on Windows we get 1848 tests vs 3161 on Travis.

XhmikosR:
Also, note that AppVeyor would still pass with snyk because they have MSYS in PATH, so in turn `patch` in PATH. Wrong but it's for convenience.

jdorfman:
If this helps interoperability between developer environments then I am all for it.

@jmervine thoughts?

jmervine:
I haven't had a chance to look at AppVeyor, but I agree with @jdorfman, that if it helps cross platform, +1.

XhmikosR:
But how about Snyk?

jmervine:
How about we leave Snyk in, but move it to it's own task and have it not run w/ `npm test`. That way we can run it optionally.

XhmikosR:
And prepublish seems to run on npm install.

jmervine:
Yeah, then just kill the `republish` entry as well. We'll have to manually run `snyk` and `snyk-protect` when we need/want it.

PR #763, opened 2016-08-31
Move jQuery code in jquery-pack.js.

XhmikosR:
This way, we can async load the vanilla JavaScript code.

/CC @jmervine

### DIFF
--- a/config/_config.yml
+++ b/config/_config.yml
@@ -16,6 +16,8 @@ javascripts:
     sri: sha384-uZ2OkdfFeYRaX81T8ZKlaMt5UJ7DkMwO8z95CYTpAtu9t6W7FaHvsa3xvalfNrnn
   main:
     uri: /assets/js/main.js
+  jquery_pack:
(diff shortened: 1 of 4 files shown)


jdorfman:
I want to discuss this first

XhmikosR:
What exactly you want to discuss?

XhmikosR:
@jdorfman: ping or we should merge this since it doesn't have any downside, only improvement.

jmervine:
I would prefer to load jquery off a cdn, is there a way we can load this off https://code.jquery.com/ instead of including it in our repo?

XhmikosR:
This isn't jQuery. I simply split our JS code.

jmervine:
😊 I see, my bad. Okay then, 👍 if @jdorfman is good with it.

XhmikosR:
@jdorfman: ping

XhmikosR:
I'm gonna merge this since I need it for my other changes.

PR #769, opened 2016-09-02
Fix public functional tests on Windows

### LINKED ISSUE #766: Different number of tests on Windows
So, while setting up AppVeyor I noticed that on Windows the same command `node make travis`, runs 1840 tests instead of 3153 tests on Travis.

/CC @jmervine for any ideas

jmervine:
Fixes #766

@XhmikosR This is better, but still not perfect. The primary issue was due to splitting out the directory using `/public/` where it needs to be `\\public\\` on windows. 

That said, the counts still don't match b/c `fs-walk` doesn't support symlinks on windows. I'm going to continue looking for a solution for symlinked files, but this is good start.

### DIFF
--- a/tests/functional_test.js
+++ b/tests/functional_test.js
@@ -90,7 +90,6 @@ function assertHeader(uri, header) {
     }
 }
 
-// bootswatch
 describe('functional', function () {
     describe('bootstrap', function () {
         config.bootstrap.forEach(function (self) {
@@ -215,7 +214,7 @@ describe('functional', function () {
         var publicURIs = [];
 
         walk.filesSync(path.join(__dirname, '..', 'public'), function (base, name) {
-            var root = base.split('/public/')[1];
+            var root = process.platform === 'win32' ? base.split('\\public\\')[1] : base.split('/public/')[1];
 
             // ensure file is in whitelisted directory
             if (typeof root === 'undefined' || whitelist.indexOf(root.split(path.sep)[0]) === -1) {
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -16,7 +16,7 @@ build: off
 test_script:
   - node -v
   - npm -v
-  - node make travis
+  - node make appveyor
 
 matrix:
   fast_finish: true
--- a/make.js
+++ b/make.js
@@ -91,6 +91,13 @@ target.travis = function () {
     target.suite();
 };
 
+target.appveyor = function () {
+    target.lint();
+
+    // without functional tests
+    assertExec(MOCHA + MOCHA_OPTS + ' -i -g "functional" -R dot ./tests/');
+};
+
 target['wp-plugin'] = function () {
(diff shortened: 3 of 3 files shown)


tests/functional_test.js
```diff
@@ -232,7 +231,7 @@ describe('functional', function () {
             }
 
             publicURIs.push(uri);
-        });
+	});
```
XhmikosR: This seems a mistake.

jmervine: Yeah, was working in windows and I don't have my vim config there to adjust whitespace.

jmervine: Fixed

tests/functional_test.js
```diff
@@ -215,15 +214,15 @@ describe('functional', function () {
...
+            var root = base.split(path.normalize('/public/'))[1];
 
             // ensure file is in whitelisted directory
             if (typeof root === 'undefined' || whitelist.indexOf(root.split(path.sep)[0]) === -1) {
                 return;
             }
 
             var domain = helpers.domainCheck('https://maxcdn.bootstrapcdn.com/');
-            var uri = domain + root + '/' + name;
+            var uri = helpers.normalizeURL(domain + root + '/' + name);
```
XhmikosR: Can't we use node.js url for this? Or `path.posix`.

jmervine: Briefly played with both and it doesn't look like it, at least not with an equal amount of massaging.

jmervine:
Perhaps, the more I'm looking in to this, I'm thinking it's not worth the effort to get this working in windows given that the issue is only with functional tests at this time. I've updated `appvayor.yml` to skip functional tests.

XhmikosR:
I will look at it later.

I want the same tests so I'll try to fix it myself.

jmervine:
@XhmikosR Sounds good. Feel free to muck with this branch at will. =D

jmervine:
@XhmikosR the more I think about it, the more I'm thinking I don't want to run functional tests on AppVayor. Functional tests hit maxcdn.bootstrapcdn.com several thousand times. Running them 4 times for each PR and twice for each commit seems a bit excessive.

@jdorfman Thoughts?

XhmikosR:
@jmervine: I don't mind us skipping functional tests on the CI level, but tests should work the same when run manually.

Please check the last patch. Locally I get `2787 passing (20m)` but as you can see it's super slow and still I don't get the same amount of tests.

I reverted the appveyor change so that we can see the results; I will redo the change before merging.

jmervine:
@XhmikosR Are you happy with this now? If so, feel free to merge.

XhmikosR:
@jmervine: still doesn't work locally the same as on Unix.

BTW, after this is sorted, I will make AppVeyor skip functional tests again since it makes sense. I just want to be able to test everything on Windows, locally.

jmervine:
I have no objections. Windows isn't really isn't in my world though.

PR #772, opened 2016-09-03
Move inline JS

XhmikosR:


XhmikosR:
I move GA track events too, but I'm not sure it's right or if it even works.

jmervine:
+1, what I was thinking

PR #783, opened 2016-10-25
Develop -> Master

jmervine:
@XhmikosR @jdorfman I rebased master to develop and this looks clean.

Additionally https://dev.bootstrapcdn.com (and most importantly) looks good to me. I'd like to see a 👍 from both of you before merged. 

I'm going to be away for the next few hours, so once both of you have given it a 👍, please feel free to merge, otherwise, I'll do it when I return.

Cheers!


XhmikosR:
LGTM.

Just a side note, the rebase thing is a one-time thing and we should never do it again as it messes up things.

PR #785, opened 2016-10-26
Switch to Node.js 6.

XhmikosR:


jmervine:
Production is currently using `v4.6.1`. To fix this, `package.json` also need to to reference `6.x` [here](https://github.com/MaxCDN/bootstrap-cdn/blob/xmr-node-6/package.json#L47).

XhmikosR:
Done.

PR #798, opened 2016-12-07
Add back serve-favicon.

XhmikosR:
Mostly for `/favicon.ico` to work.

@jmervine: this should be doable with a couple of lines in app.js but my express.js knowledge is not very mature.

What do you think we should do?

### DIFF
--- a/app.js
+++ b/app.js
@@ -13,6 +13,7 @@ var sitemap     = require('express-sitemap');
 var app = express();
 
 // middleware
+var favicon      = require('serve-favicon');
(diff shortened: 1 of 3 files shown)


jmervine:
This could be done by dropping the favicon in the public directory and just letting it be picked up as static -- but then we lose the specific cache control. We could also create a custom route that serves it up with a custom cache header.

That said, I think `serve-favicon` -- how you implemented it -- is cleaner, clearer and more abstracted, so...

👍

PR #802, opened 2016-12-07
Move Google Analytics track events to main.js.

XhmikosR:
This is something 1) I haven't used before myself and 2) most important, I can't check it myself since I don't have access to the GA account.

That being said, this is the last inline JavaScript we use and after that we can remove `unsafe-inline` from CSP `script-src`. So, we should definitely try and get this sorted.

### DIFF
--- a/app.js
+++ b/app.js
@@ -102,7 +102,6 @@ app.use(helmet.contentSecurityPolicy({
         defaultSrc: ['\'none\''],
         scriptSrc: [
             '\'self\'',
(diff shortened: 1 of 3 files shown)


jmervine:
Same on both accounts, so can't help you here. Code changes look clean though, so 👍 from that perspective.

@jdorfman can you help us with the GA stuff?

jdorfman:
:shipit:

PR #804, opened 2016-12-09
Inline style.css and main.js.

XhmikosR:
The reason for this is that these files are very small.

Older browsers will use `'unsafe-inline'`, newer ones will use the supplied `nonce`.

### DIFF
--- a/app.js
+++ b/app.js
@@ -2,28 +2,28 @@
 
 var env = process.env.NODE_ENV || 'development';
 
-var path        = require('path');
-var fs          = require('fs');
-var http        = require('http');
-var express     = require('express');
-var yaml        = require('js-yaml');
-var compression = require('compression');
-var sitemap     = require('express-sitemap');
-
-var app = express();
+var path         = require('path');
+var fs           = require('fs');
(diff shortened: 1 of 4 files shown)


XhmikosR:
OK, I managed to fix it. I think it's OK now. The random string is generated on app start.

/CC @ScottHelme for a quick review.

jmervine:
👍

XhmikosR:
@jmervine: only issue is that those files are minified now regardless of production or not.

Is there any way to make this conditional?

Finally, when I run [CSP Evaluator](https://csp-evaluator.withgoogle.com/), I notice it says `'nonce-a267e5b1-272e-4c39-9aca-aacc5395b6ad'  Nonces should only use the base64 charset.`. So, I wonder, should I base64 the nonce?

XhmikosR:
Oh, and `cdn.syndication.twimg.com is known to host JSONP endpoints which allow to bypass this CSP.`

XhmikosR:
OK, I believe I sorted all issues, apart from the base64 info.

PR #807, opened 2016-12-11
Tweak main.js.

XhmikosR:
* Fix CSS selector syntax in `selectInputContent`; Safari 10 threw a syntax error
* Remove unneeded check for return value of `querySelectorAll`
* Remove calls to `window.attachEvent`; IE8 isn't supported
* Name our functions

@jmervine or @jdorfman: I don't have a Mac machine available, so I only managed to test before this patch. If someone with the latest Safari could verify that there's no JS error that would be great.

Also, regarding Internet Explorer, IE 10 and newer is supported. I could try to support IE 9 but I don't think it's worth it.

### DIFF
--- a/public/assets/js/main.js
+++ b/public/assets/js/main.js
@@ -1,10 +1,11 @@
-(function() {
+(function mainJS() {
     'use strict';
 
-    (function () {
-        var el = document.querySelectorAll('.input-group-btn > button');
+    (function toggleInputCaret() {
+        var selector = '.input-group-btn > button';
+        var el = document.querySelectorAll(selector);
 
-        function toggleCode (index) {
+        function toggleCode(index) {
             var name = el[index].getAttribute('data-el');
             var btnIcon = el[index].querySelector('span');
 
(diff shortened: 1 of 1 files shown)


XhmikosR:
`https://www.bootstrapcdn.com`: https://www.browserstack.com/screenshots/0939a3fad1f2165434f8fb8cc667036757d4ae90

`https://bootstrapcdn-dev-pr-807.herokuapp.com/` https://www.browserstack.com/screenshots/3ac943961763cb95e7cc57d38576ef81357b4494

So it seems more old browsers, like Google Nexus 6, had issues with the JS error.

public/assets/js/main.js
```diff
@@ -1,10 +1,11 @@
-(function() {
+(function mainJS() {
```
jmervine: Always did prefer to name anonymous function. =P

XhmikosR: There is an ESLint for this, it's just that it results in so many warnings.

I'll make a PR some day with some tweaks and ping you.

PR #808, opened 2016-12-11
Remove npm scripts.

XhmikosR:
No reason in duplicating these since we have the `node make` targets.


XhmikosR:
I don't mind either, but duplication is always bad... For example, we have `"test-nc": "node make test-nc"` which doesn't exist.

That is why I prefer to have everything in one place, using one way: the way I also use myself...

jmervine:
👍 🚢

PR #811, opened 2016-12-12
Add social-related meta tags.

XhmikosR:
This is work in progress, do **not** merge yet.

This is how it will look like: <http://iframely.com/debug?uri=https%3A%2F%2Fbootstrapcdn-dev-pr-811.herokuapp.com%2F>

### DIFF
--- a/routes/index.js
+++ b/routes/index.js
@@ -15,6 +15,8 @@ function appendLocals(req, res) {
 
     res.locals.fullUrl = proto + '://' + req.hostname + req.path;
 
+    res.locals.siteUrl = proto + '://' + req.hostname;
+
     res.locals.theme = req.query.theme;
 
     res.locals.displayTitle = function (pageTitle) {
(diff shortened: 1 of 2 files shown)


XhmikosR:
This is how it will look like  http://iframely.com/debug?uri=https%3A%2F%2Fbootstrapcdn-dev-pr-811.herokuapp.com%2F

XhmikosR:
@jmervine: my only concern about this is if we should use a hardcoded `siteUrl`. I mean, it doesn't make sense to be different let alone I'm not sure if it can hurt SEO, which probably will.

jdorfman:
@XhmikosR it can't hurt SEO.

XhmikosR:
Well it certainly can't help either since both dev and normal site are
being crawled... But if you both agree, we can go with the current patch
and tweak it if needed.

jmervine:
I don't think there's an issue leaving it dynamic for off domains, although not strictly necessary. Were it me, I'd probably leave dynamic, just because it's already that way.

PR #815, opened 2016-12-12
Production release

XhmikosR:
Not sure if we want to wait a little more to merge the other 2 PRs. Probably a good idea, but we can merge them, and this PR will be updated automatically.


app.js
```diff
@@ -74,82 +74,110 @@ app.use(function (req, res, next) {
...
+}));
 
+app.use(helmet.hsts({
+    force: true,
+    includeSubdomains: true,
+    maxAge: 10886400,
+    preload: true
+}));
+
+app.use(helmet.contentSecurityPolicy({
```
jmervine: This down the road we should pull this out into it's own module... getting it out of app.js.

jmervine:
@XhmikosR Feel free to merge this when you're ready. Once done, it takes ~10 minutes or so for the release to be live.

app.js
XhmikosR: I agree, app.js is getting big. 👍

PR #818, opened 2016-12-12
Assorted code cleanup

XhmikosR:
I tested this, seems to work fine, but please have a close look in case I missed something.

### DIFF
--- a/lib/helpers.js
+++ b/lib/helpers.js
@@ -16,7 +16,7 @@ function sriDigest(file, fromString) {
 }
 
 function selectedTheme(config, selected) {
-    if (typeof selected === 'undefined' || selected === 'undefined') {
+    if (typeof selected === 'undefined') {
         return config.theme;
     }
 
@@ -25,8 +25,9 @@ function selectedTheme(config, selected) {
 }
 
 function getTheme(config, selected) {
-    selected = selectedTheme(config, selected);
     var themes = config.bootswatch.themes;
+
+    selected = selectedTheme(config, selected);
 
     return {
         uri: config.bootswatch.bootstrap
--- a/tests/functional_test.js
+++ b/tests/functional_test.js
@@ -31,14 +31,13 @@ var expectedHeaders = {
     'last-modified': undefined,
     'x-cache': undefined,
 
-    // 'x-amz-server-side-encryption': 'AES256',
     'accept-ranges': undefined,
     'access-control-allow-origin': '*',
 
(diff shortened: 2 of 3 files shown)


lib/helpers.js
```diff
@@ -16,7 +16,7 @@ function sriDigest(file, fromString) {
 }
 
 function selectedTheme(config, selected) {
-    if (typeof selected === 'undefined' || selected === 'undefined') {
```
jmervine: I'm pretty sure there was a case where I was returning `'undefined'` as a string for selected. Might be long gone now though.

tests/functional_test.js
```diff
@@ -222,8 +214,7 @@ describe('functional', function () {
                 return;
             }
 
-            var domain = helpers.domainCheck('https://maxcdn.bootstrapcdn.com/');
-            var uri = domain + root + '/' + name;
+            var uri = 'https://maxcdn.bootstrapcdn.com/' + root + '/' + name;
```
jmervine: This wasn't hardcoded, because I wanted to be able to test S3 separately when run locally, to debug inconsistencies.

lib/helpers.js
XhmikosR: I can't find any relevant code anywhere so it was probably in the past.

tests/functional_test.js
jmervine: I'm on the fence, but I think I want to keep it. This is how I validate uploads to S3 before purging cache.

jmervine: If/when we move off S3, I'd want to nuke this for sure.

XhmikosR: All right, I'll see what I can keep from these patches, otherwise I'll close the PR.

lib/helpers.js
jmervine: Okay, then this is probably a safe and a good find at that. =P

XhmikosR:
@jmervine; please check one last time.

lib/helpers.js
XhmikosR: I think I found the relevant part https://github.com/MaxCDN/bootstrap-cdn/blob/develop/views/_partials/header.pug#L1

This must have changed in pug and they don't pass an empty string IIRC. Is there any way you could check #818 console log for any errors while browsing the site and switching themes?

jmervine: Yeah, I'll need to do that later this afternoon though.

PR #819, opened 2016-12-13
Add pug-lint

XhmikosR:


jmervine:
@XhmikosR I'm going to let you resolve these conflicts, don't want to mess anything update.

PR #824, opened 2016-12-16
Add `rel='noopener noreferrer'` for all of our `target="_blank"` links.

XhmikosR:
See https://www.jitbit.com/alexblog/256-targetblank---the-most-underestimated-vulnerability-ever/ for more info behind this.

Also, use `target="_blank"` consistently for all outbound links.


XhmikosR:
I'm thinking about this more. Maybe we shouldn't have `noreferrer` for the links we want to track?

I don't have access to the Google Analytics account so I can't verify this myself.

/CC @jdorfman

PR #826, opened 2016-12-16
**DO NOT MERGE** Testing loadjs loader.

XhmikosR:
**DO NOT MERGE**

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,2 +1,3 @@
 /.git/
 /node_modules/
+/public/assets/js/vendor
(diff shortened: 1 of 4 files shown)


jmervine:
What does this add? Link preloading like Turbolinks?

Also it'd be nice to minify loader.js

XhmikosR:
The whole idea was to reduce the number of files needed to load the page.

I'm not sure if I'll go with this solution or not. It was mostly an experiment which works BTW.

As for load.js being unminified, we minify it for production. Unless you mean have it in the repo minified.

PR #827, opened 2016-12-17
Merge Bootlint scripts and convert it to JS.

XhmikosR:
**WORK IN PROGRESS, DO NOT MERGE!**

@jmervine: there are 2 issues at this point:

1. `ln` complains that the file doesn't exist which is right ("ln: Source file does not exist"); not sure how else I can create the symlink
2. uglify fails the first time; it's like it's running before `http.get` is done


jmervine:
I'll take a look later this afternoon.

XhmikosR:
Thanks. Keep also in mind that the `ln` error might be because I'm on Windows and there's poor support for it. So we make it work on *nix, it's OK for me.

scripts/update_bootlint.js
```diff
@@ -0,0 +1,69 @@
...
+https.get('https://raw.githubusercontent.com/twbs/bootlint/v' + version + '/dist/browser/bootlint.js', function (res) {
+    var statusCode = res.statusCode;
+    var error = { };
+    var file = fs.createWriteStream(path.join(bootlintDir, 'bootlint.js'));
+
+    if (statusCode !== 200) {
+        error = new Error('Request Failed.\nStatus Code: ' + statusCode);
+    }
+
+    if (error) {
```
jmervine: I believe this would always return true, a few tests on the node console...

```node
jmervine@jmervine-ltm ~ $ node
> var error = {};
undefined
> !!(error);
true
> var error;
undefined
> !!(error);
true
> var error = undefined;
undefined
> !!(error);
false
```

`error` should be set to `undefined` on line 37.

```node
var error = undefined;
```

scripts/update_bootlint.js
```diff
@@ -0,0 +1,69 @@
...
+var bootlintDir = path.join(basedir, 'public', 'bootlint', version);
+var UGLIFYJS    = path.join(basedir, 'node_modules/.bin/uglifyjs');
+
+if (!version) {
+    echo('Valid Bootlint version required.');
+    process.exit(1);
+}
+
+// strip leading 'v' if present
+version = version.replace(/^v/, '$1');
```
jmervine: Is `$1` correct here? Shouldn't it be `version.replace(/^v/, '')`? Perhaps this is a shell js thing I'm not aware of?

scripts/update_bootlint.js
jmervine: As an aside, I would simplify error checking by putting it all in a single conditional and remove the need for `error` all together...

```node
if (statusCode !== 200) {
   console.log(new Error('Request Failed.\nStatus Code: ' + statusCode).message);
   res.resume();
   return;
}
```

But totally up to you, if you prefer to keep them separate.

scripts/update_bootlint.js
```diff
@@ -0,0 +1,69 @@
...
+        console.log('Got error: ' + e.message);
+    });
+});
+
+exec(UGLIFYJS + ' bootlint.js -o bootlint.min.js ' +
+  '--compress --source-map bootlint.min.js.map ' +
+  '--comments "/(?:^!|@(?:license|preserve|cc_on))/"');
+
+cd('..');
+ln('-sf', 'latest', version);
```
jmervine: If you're not 100% sure about `ln` support here in windows, perhaps check for `win32` and output a note to manually add a symlink here?

scripts/update_bootlint.js
```diff
@@ -0,0 +1,69 @@
...
+
+    res.on('end', function () {
+        file.close();
+    })
+    .on('error', function (e) {
+        console.log('Got error: ' + e.message);
+    });
+});
+
+exec(UGLIFYJS + ' bootlint.js -o bootlint.min.js ' +
```
jmervine: Move this inside the `https.get` callback and it should exec in the proper order. In fact everything should be inside that callback.

XhmikosR:
@jmervine: thanks for the very useful comments!

1. I refactored the error check like you suggested; everything's good there now.
2. I worked around the Windows symlink issue by just writing the version as simple file; please check out the branch to see if it looks like a symlink on your platform. If not, maybe I should try `fs.link` or something.
3. There's still the issue with uglify which I can't solve.

jmervine:
👍 on the windows symlink issue, I totally forgot that's how github handles symlinks

RE uglify, I'll try to look later today.

jmervine:
Never got a chance to look at this. I can try next week. Apologies.

jmervine:
@XhmikosR I **finally** looked at this. Sorry it took so long. Made some tweaks. LMK what you think. Happy to collaborate, revert and revisit, etc.

Cheers!

XhmikosR:
@jmervine: thanks!

A couple of things:

1. we need to pass a flag to uglify to make `sourceMappingURL` relative because currently I get `//# sourceMappingURL=C:\Users\xmr\Desktop\bootstrap-cdn\public\bootlint\0.14.2\bootlint.min.js.map`
2. did you try it on your machine after removing the symlink and 0.14.2 folder? Because on Windows things are messy with symlinks and I can't trust the results.

XhmikosR:
I believe I fixed source maps. So now only thing left is to verify the symlink  is valid on nix systems.

XhmikosR:
Oh, and we need to tweak the order of version check since as it is now, `replace` will throw if no version is specified.

jmervine:
Should we work to get 847 merged and the ES6'ify this, or the otherway, merge this, rebase to 847 and ES6 it there?

scripts/update_bootlint.js
```diff
@@ -0,0 +1,71 @@
...
+    echo('Valid Bootlint version required.');
+    process.exit(1);
+}
+
+if (test('-d', bootlintDir)) {
+    echo('Bootlint version already found.');
+    process.exit(1);
+}
+
+https.get('https://raw.githubusercontent.com/twbs/bootlint/v' + version + '/dist/browser/bootlint.js', function (res) {
```
jmervine: Since this is more of a script, might this be a good use-case for `await https.get(uri)`?

XhmikosR: That I don't know, I haven't used it either yet...

XhmikosR:
Your call, I don't mind rebasing either one.

jmervine:
I'd say merge #847 if you feel it's ready (I do) and then rebase this one. Probably cleaner.

jdorfman:
@XhmikosR close?

XhmikosR:
It's not done yet either.

PR #828, opened 2016-12-18
Update CSP directives.

XhmikosR:
**Do not merge, still investigating.**

@ScottHelme: is there any better way doing this? Using `*`, seems too broad.


XhmikosR:
@ScottHelme: is there any better way doing this? Using `*`, seems too broad.

ScottHelme:
If you have a broad set of subdomains and don't want to bloat the header then this is the only way. I guess given the twimg.com they don't have any 3rd party managed subdomains (probably?) so it's not too broad.

XhmikosR:
@ScottHelme: thanks for the insight!

PR #829, opened 2016-12-18
**WIP** Try out InstaClick.

XhmikosR:
https://github.com/dieulot/instantclick


jmervine:
This is similar to Rail's Turbolinks. I'm 👍 adding this if it works smoothly. 

I would caution though, that if we ever get in to adding form handling / form submit features (which I can't see why we would), we'll need to revisit this and selectively disable it around that.

XhmikosR:
I'm having issues at this point and I need to look closer at the docs. Also, I'm not sure if there's a newer lib available that does the same thing or not.

This is mostly an experiment and I had to open the PR to have Heroku build the branch. If there's another way, let me know.

jmervine:
FWIW, https://github.com/turbolinks/turbolinks is tried an true, should do basically the same thing and works outside of rails.

XhmikosR:
Thanks, looks good too. But it seems a lot bigger...

I'll keep this branch open until I have some more time to look at it.

jmervine:
Yeah, turbolinks is huge -- as all things Rails is huge =P

PR #823, opened 2016-12-16
Production Release

jmervine:
@XhmikosR @jdorfman

PR #831, opened 2016-12-20
Update CSP directive.

XhmikosR:
For some reason, Safari 10 complains about not being able to load the font files from the CDN, even though it shouldn't even try to download them in first place.

This should fix those errors.

### DIFF
--- a/app.js
+++ b/app.js
@@ -136,9 +136,9 @@ app.use(helmet.contentSecurityPolicy({
         ],
         fontSrc: [
             '\'self\'',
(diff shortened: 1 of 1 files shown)


jmervine:
Is the CI error due to #832 ? If so, we should disable html validation until it's resolved.

XhmikosR:
Actually, it is sort of related to #823. It's because html-validator is using the online version of the validator which gets updated from the master branch and thus changes come faster.

I've opened an issue, see https://github.com/validator/validator/issues/433

PR #832, opened 2016-12-20
Switch to html-validator-cli.

### LINKED ISSUE #693: Run HTML validation for all pages

XhmikosR:
- [ ] refactor make.js to avoid code duplication
- [x] We need to run the validator for each page and not all pages at once since this doesn't work

---

~~@jmervine: this is sort of urgent in the sense that due to the html-validator recent updates, tests fail. So, we might as well try and switch to this solution so that we check all pages.~~

Feel free to push here and I will rebase/clean up the branch before merging, assuming we get everything to work ofc :)

Fixes #693.

### DIFF
--- a/make.js
+++ b/make.js
@@ -137,46 +137,41 @@ target.bootlint = () => {
 
     const outputs = [];
 
-    // sleep
-    setTimeout(() => {
-        echo('------------------------------------------------');
-        async.eachSeries(pages, (page, callback) => {
-            const url = `http://localhost:${port}/${page}${page === '' ? '' : '/'}`;
+    echo('------------------------------------------------');
+    async.eachSeries(pages, (page, callback) => {
+        const url = `http://localhost:${port}/${page}${page === '' ? '' : '/'}`;
 
-            if (page !== '') {
-                page += '_';
-            }
+        if (page !== '') {
+            page += '_';
+        }
 
-            const output = path.join(__dirname, `${page}lint.html`);
-            const file = fs.createWriteStream(output);
+        const output = path.join(__dirname, `${page}lint.html`);
+        const file = fs.createWriteStream(output);
 
-            // okay, not really curl, but it communicates
-            echo(`+ curl ${url} > ${output}`);
+        // okay, not really curl, but it communicates
+        echo(`+ curl ${url} > ${output}`);
 
-            http.get(url, (response) => {
-                response.pipe(file);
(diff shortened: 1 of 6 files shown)


jmervine:
I wouldn't mind looking at following the `rake` pattern of creating a `lib/tasks` dir, where we move things to modules for import in to `main.js`. Some separation might be helpful.

XhmikosR:
@jmervine: agreed, but I'd leave that for when we are done with most things and only refactoring is left.

That being said, if you want to tackle this first that's fine by me :)

XhmikosR:
@jmervine: when you have some time, I'd really appreciate if you could tackle the two remaining issues in my original post above.

XhmikosR:
From a quick look, it seems the error we get is indeed valid; the end script tag is missing so something doesn't finish before the validator runs.

jdorfman:
close?

XhmikosR:
No, it's not done.

XhmikosR:
@jmervine: can you please review? It seems I got it to work right now, it's just that I'm running one request/file per time.

make.js
```diff
@@ -175,6 +179,78 @@ target.bootlint = () => {
     }, 2000);
 };
 
+target.htmllint = () => {
+    const port = 3081;
+
+    env.PORT = port;
```
jmervine: It's failing CI because the port your assigning isn't being applied on startup. `process.env.PORT` should work though, as the first line of the `start` target is `const env = process.env`.

make.js
```diff
@@ -175,6 +179,78 @@ target.bootlint = () => {
...
+    env.NODE_ENV = 'development';
```
jmervine: Same with this, try `process.env.NODE_ENV = 'development';`.

XhmikosR: I think something else is wrong. Running `node make htmllint` works fine; it breaks when bootlint is run before the htmllint target. It's like the next target starts while the old one hasn't finished yet.

jmervine:
Looks like that didn't work, perhaps either use the default port, or modify the start task to take an optional param?

XhmikosR:
I have no idea how to solve this :/

I still believe the issue is that the targets are somehow overlapping with each other so when `htmllint` starts, bootlint hasn't finished yet.

I also tried updating shelljs to the latest version which works fine, until an error occurs; then it throws. Si suspect it's our `assertExec` function.

PR #834, opened 2016-12-21
Ignore the new HTML validation error about `link rel=apple-touch-icon`.

XhmikosR:


tests/test_helper.js
```diff
@@ -108,11 +108,18 @@ function assertValidHTML(response, done) {
                 return false;
             })
             .filter(function(e) {
-                console.log('>> e:', e);
-                if (e.match(/^Error: Attribute.+color.+not allowed on element.+link.+at this point./)) {
+                var ignoreRe = /^Error: Attribute.+color.+not allowed on element.+link.+at this point./;
+                var ignoreRe2 = /^Error: A.+link.+element with a.+sizes.+attribute must have a.+rel.+attribute that contains the value.+icon.+./;
```
XhmikosR: I thought about it too, but I was hoping we wouldn't need to do that. I can refactor the code, sure, but you know, we better be HTML5 valid at all times unless we really need not to.

Those 2 errors are reported upstream to https://github.com/w3c/html; still waiting for a solution.

jmervine: Perhaps we go as is, and then refactor if/when we need to add more. Sound good?

XhmikosR: Let me have a look again later. I see this is merged https://github.com/w3c/html/pull/757 so the validator needs to be updated.

PR #835, opened 2016-12-21
Remove HTML caching since we are using nonces.

XhmikosR:
@ScottHelme: I tried finding any references about this without a lot of success. So, from what I understand, the HTML shouldn't be cached when nonce's are being used. Is this correct?

Thanks!

### DIFF
--- a/app.js
+++ b/app.js
@@ -66,17 +66,13 @@ app.use(serveStatic(path.join(__dirname, 'public'), {
 }));
 
 app.use(function (req, res, next) {
-    var oneHourToSec = 60 * 60;
-    var oneHourToMilliSec = 60 * 60 * 1000;
-
     // make config available in routes
     req.config = config;
 
     // custom headers
     res.setHeader('X-Powered-By', 'MaxCDN');
     res.setHeader('X-Hello-Human', 'You must be bored. You should work for us. Email jdorfman+theheader@maxcdn.com or @jdorfman on Twitter.');
-    res.setHeader('Cache-Control', 'public, max-age=' + oneHourToSec);
-    res.setHeader('Expires', new Date(Date.now() + oneHourToMilliSec).toUTCString());
+    res.setHeader('Cache-Control', 'public, max-age=0');
     res.setHeader('Last-Modified', new Date().toUTCString());
     res.setHeader('Accept-Ranges', 'bytes');
 


ScottHelme:
Hmm I'm not sure on this one, good question. I use Nginx as a caching proxy and I cache the page prior to substitution of the nonce marker so even though the page is cached I write in a random nonce value each load. I hadn't though about client side though... I will ask around :-)

XhmikosR:
Thanks!

jmervine:
Removing [this line](https://github.com/MaxCDN/bootstrap-cdn/blob/develop/tests/functional_test.js#L19) should fix the tests.

XhmikosR:
@jmervine: I didn't notice due to CI failing for other reasons. I will take care of it if we go with this solution (I'm waiting for ScottHelme's feedback).

XhmikosR:
@jmervine: FYI, that line is for the CDN assets and not for the website's HTML files. So, there are no failures. The previous one was due to the HTML validation error.

ScottHelme:
I spoke to @mikewest and he made the great point that hashes would be a better fit here. I guess whether or not that's a viable solution will depend on how many assets you'd need to hash and include in the policy. Nonces will not play nice with client side caching.

XhmikosR:
Isn't the current patch a viable solution? Granted that we don't mind **not** caching the HTML files.

I don't know the difference between hashes and nonce so I don't know the pros and cons. I'll need to read up on that later.

ScottHelme:
2 quite different methods, at a high level: 

Hashes: insert the hash of the script/style into the CSP like so 'sha256-[hash]'. If you have 10 scripts and/or styles though, you end up with that many hashes. The browser will hash each script and style looking for a match. If it matches one, it's allowed.
https://scotthelme.co.uk/csp-cheat-sheet/#hashes

Nonces: even if you have 100 scripts and styles they can all share the same nonce, so it's less of a burden in this regard, though if an attacker can get the nonce, they can fully bypass the CSP. 
https://scotthelme.co.uk/csp-cheat-sheet/#nonces

XhmikosR:
Thanks for the explanation! I had already your page in my bookmarks and forgot about it :p

Now, I believe the current approach is a good compromise for our case. It's less of a hassle to set up, less headers and I don't think we care a lot about HTML caching.

@jmervine: what are your thoughts on the matter?

jmervine:
Hashing seems safer overall and we don't have that many scripts included for me to worry about their being a burden. 

That said, I agree that we don't care much about HTML caching, so the currently implemented solutions seems acceptable.

PR #836, opened 2016-12-22
Minor Jade consistency tweaks.

XhmikosR:


views/bootswatch.pug
```diff
@@ -3,10 +3,11 @@ extends layout
 block content
     h2 Bootswatch
     .well
-        - for (var i = 0; i < config.bootswatch.themes.length; i++) {
+        - for (var i = 0, len = config.bootswatch.themes.length; i < len; i++) {
```
jmervine: This seems unnecessarily verbose. Or am I missing something?

XhmikosR: Just a matter of habit :) I can revert it if you want.

PR #838, opened 2016-12-22
Remove unneeded dependency serve-static.

XhmikosR:
Express.js has this built-in.

Not sure if we need to require this explicitly.


jmervine:
👍🏽 If you're sure. I remember needing it, but that may have been an older version of express.

XhmikosR:
http://expressjs.com/en/api.html#express.static

And everything seems to work so I'm gonna merge it.

PR #839, opened 2016-12-23
Use the unicode symbols when possible.

XhmikosR:


views/_partials/header.pug
```diff
@@ -27,7 +27,8 @@
                 li(class=(title === 'Integrations' ? 'active' : ''))
                     a(href='/integrations/' + themeQs) Integrations
                 li.dropdown
-                    a.dropdown-toggle(href='#', data-toggle='dropdown') Resources&nbsp;
+                    a.dropdown-toggle(href='#', data-toggle='dropdown') Resources
```
jmervine: This change didn't modify the display at all? I remember the `&nbsp;` being required, but perhaps I'm wrong.

jmervine:
One question, otherwise, 👍

views/_partials/header.pug
XhmikosR: Normally it is; I just used `|  ` instead which we use elsewhere.

PR #844, opened 2017-01-26
Add clipboard.js to snippets.

XhmikosR:
This is **NOT** ready yet; do **NOT** merge.

I need some feedback on the implementation and if anyone has any improvement suggestion, please, do let me know.


jmervine:
I love this idea. What's left to get it shipable?

jdorfman:
@jmervine good call. @XhmikosR what's the deal? This would be awesome.

XhmikosR:
Needs so tweaking. I also need to show something when finished.

On a side note, @jmervine the branches are not being built anymore?

XhmikosR:
@jdorfman @jmervine: can you check the last patch? Might need some tweaks but generally seems to work fine.

XhmikosR:
There's an issue when copying the small `input`s; I'll try to fix it.

views/alpha.pug
```diff
@@ -30,6 +30,7 @@ block content
                 .input-group-btn
                     button.btn.btn-default.btn-lg(type='button', data-el=name)
                         span.caret
+            span.help-block Click to copy
```
jmervine: I realized that there seems to be a lot of redundancy here with 

```
         .form-group
             label(for=id) Complete CSS
             .input-group.input-group-lg
                 input.form-control(id=id, type='text', value=file)
                  .input-group-btn
                      button.btn.btn-default.btn-lg(type='button', data-el=name)
                          span.caret
             span.help-block Click to copy
```

I realize that this isn't due to this PR, but am thinking we could abstract the whole section to an include. Perhaps in another PR -- which I'm happy to submit once this is merged.

public/assets/js/main.js
```diff
@@ -20,18 +22,53 @@
...
+                var clipboardSnippets = new Clipboard(this, {
+                    target: function (trigger) {
+                        return trigger;
+                    }
+                });
+
+                clipboardSnippets.on('success', function (e) {
+                    var helpBlock = {};
+
+                    if (e.trigger.parentElement.nextElementSibling &&
```
jmervine: Perhaps set `var nextSibling = e.trigger.parentElement.nextElementSibling;` to avoid repeating the verbose reference.

public/assets/js/main.js
```diff
@@ -20,18 +22,53 @@
         }
     })();
 
-    (function selectInputContent() {
+    (function selectTextCopyToClipboard() {
         var selector = 'input[type="text"]';
         var el = document.querySelectorAll(selector);
+        var origHelpBlockText = 'Click to copy';
```
jmervine: Were we to abstract the pug blocks to an include (as suggested below), it might be better to set this by reading what's currently set in the element, so that we only have to change it on place, should we decide on different messaging in the future -- not that I foresee doing that, but still.

public/assets/js/main.js
```diff
@@ -20,18 +22,53 @@
...
+
+                    helpBlock.innerHTML = 'Copied text to clipboard';
+                });
+
+            }, true);
+
+            el[i].addEventListener('blur', function(e) {
+                var helpBlock = {};
+
+                if (this.parentElement.nextElementSibling &&
```
jmervine: Same comment as above regarding setting this to a shorter reference variable.

jmervine:
@XhmikosR Some idle comments which you can take or leave, nothing I feel as critical. Looks great!

public/assets/js/main.js
XhmikosR: Yeah, that is what I had, but it ESLint annoyed me with the no-loop-func errors so I hardcoded.

Agreed though it should be read with `innerHTML`.

views/alpha.pug
XhmikosR: Agreed, I noticed it too.

XhmikosR:
@jmervine: thanks for the review! If you could help with retrieving the `span`s text that would be great.

On a side note, alternatively we could use `el.classList.contains()`; not sure if it will be cleaner or not.

jmervine:
I can take a look, I don't have much experience with client side js outside of using jquery, which you're obviously trying to avoid, but let me play in the console and see what I can do.

XhmikosR:
It's just https://developer.mozilla.org/en-US/docs/Web/API/Element/innerHTML :)

But because we are in a `for` loop, you know :p

I plan to switch to ES5 eventually.

jmervine:
```
> document.getElementsByClassName('help-block')[0].innerHTML
"Click to copy"
```

XhmikosR:
Wait, you want to just get the value of the first span or each one of them? Because, sure they are the same, but not sure what it can happen in the future.

jmervine:
If we're building all the spans with an include (eventually) we can be sure they'll all be the same. Thus getting the first one would probably suffice.

XhmikosR:
Then in that case I can just use `document.querySelector`.

jmervine:
I'd probably keep it hardcoded until we abstract that template block though, because until then it doesn't really matter. Should we change that text, we'd need to change it in multiple places as is.

XhmikosR:
All right. Then I'll leave things as they are.

jmervine:
Sounds good. Once we merge this, I'll take a stab at moving those blocks out to an include and update the js accordingly.

jmervine:
👍 🚢

XhmikosR:
One final thought is if we should make JS external again; it's 12.4 KB minified. But we lack http/2 so all this would make things worse. On the hand the JS would be cached. Oh, plus last time I tried, I couldn't make the middleware which minifies files to work.

jmervine:
Given that http/2 isn't on the horizon near term, we should make the choices that are best of http/1 at this time.

RE the minify middleware, I can try and look at that this evening, but I can't commit to it.

XhmikosR:
Why did you merge it @jdorfman?

I could do it myself... I had things I wanted to investigate first.

jdorfman:
@XhmikosR shit my bad I though it was waiting on me.

XhmikosR:
No big deal I guess, it's on dev only.

We'll need  see if we are going to split JS to an external file but make sure we minify it.

jdorfman:
@XhmikosR holy shit that click to copy is cool! Great implementation

XhmikosR:
So, you think it's better with click instead of a button, right?

jdorfman:
Yeah it is innovative. I never seen anything like it

jdorfman:
@XhmikosR lets :shipit:

XhmikosR:
Not yet I want to try making Javascript external.

PR #848, opened 2017-05-08
Add missing solar theme

bostondevin:
This one somehow was missing


XhmikosR:
Thanks but you need to update config.yml too.

bostondevin:
Where are you getting the sri value from? Looks like a ssh key or something - I don't know how to generate that

XhmikosR:
You can do it manually or use srihash.org

bostondevin:
hmmm srihash.org wants a URL, so the file would have to exist there already

e.g.
https://maxcdn.bootstrapcdn.com/bootswatch/3.3.7/solar/bootstrap.min.css

XhmikosR:
Right. Then use scripts/integrity.js.

bostondevin:
not sure what u mean - this is more complicated than one would think!

XhmikosR:
I'll have a look and add it myself later then.

bostondevin:
that would be awesome - thanks! Never heard of sri's before

XhmikosR:
Do you know when solar was added?

bostondevin:
I can't quite place it. It's on their website as being in this latest version - you might have needed to build the project or something to create the new files - i dunno

jdorfman:
Thanks @bostondevin !

PR #857, opened 2017-07-14
Use Bootstrap's collapse for wells.

XhmikosR:


XhmikosR:
It's not perfect since the toggle button action happens instantly and the collapse has a delay but it's better than before.

Any suggestions/patches welcome.

XhmikosR:
/CC @Johann-S just in case you have time and any improvement ideas.

Johann-S:
Maybe that's related to : https://github.com/twbs/bootstrap/pull/20870 🤔

public/assets/js/main.js
```diff
@@ -8,11 +8,9 @@
         var el = document.querySelectorAll(selector);
 
         function toggleCode(index) {
-            var name = el[index].getAttribute('data-el');
             var btnIcon = el[index].querySelector('span');
 
             el[index].addEventListener('click', function() {
-                document.getElementById(name).classList.toggle('hidden');
                 btnIcon.classList.toggle('caret-open');
```
Johann-S: Maybe you can do : 

```javascript
setTimeout(function () {
  btnIcon.classList.toggle('caret-open');
}, 350);
```

`350` is the transition duration of the Collapse plugin (https://github.com/twbs/bootstrap/blob/v3-dev/js/collapse.js#L35)

XhmikosR:
I haven't tried that but I will, thanks! I was trying to use transionend
but didn't spend a lot of time.

jmervine:
:+1:

PR #859, opened 2017-07-17
moving uglify-js

jmervine:
want to see if the PR app works

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,7 +22,8 @@
     "serve-favicon": "^2.3.2",
     "shelljs": "^0.5.3",
     "snyk": "^1.19.1",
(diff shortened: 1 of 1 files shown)


jmervine:
As discussed, the package-lock.json doesn't change, so this shouldn't make any different in production. But sometimes Node / NPM is weird.

XhmikosR:
The PR works indeed.

But maybe it's just a matter of triggering a new build?

jmervine:
I submitted a empty commit on develop priory to this PR to ensure that a new build was trigged, so I don't _think_ that's it, but maybe. 🤷‍♂️

PR #871, opened 2017-10-13
Update dependencies.

XhmikosR:
@jmervine: for some unknown reason to me, tests fail currently. I don't think it's related to these changes, though.

I left uglify-js to the old version so that the minified files don't change. Also, shell-js because the new version throws differently when an error occurs. We should update it eventually, but you know...


XhmikosR:
OK, it seems mocha changed their behavior and we need to add `--exit`.

@jmervine: LGTY assuming tests pass?

Issue #867, opened 2017-08-12
Remove twitter from boostrap path


XhmikosR:
At least for v4. Personally I'd remove it from everywhere but I'm not familiar how things work with S3.

/CC @jmervine @jdorfman

jmervine:
@jdorfman can we check the cdn and see if anyone is even hitting that?

XhmikosR:
Any news about this?

jmervine:
@jdorfman ^^

jdorfman:
We don't need to advertise it, but not going to remove it. Even if it is 1 hit a year, I still want that legacy use case to work.

XhmikosR:
And why can't you redirect the old path?

jdorfman:
S3 doesn't support symlinks. They do have a hack but it makes the http request 301 which is unacceptable.

https://stackoverflow.com/a/29462540/1258484

jdorfman:
@jmervine can we just adjust the tests to ignore `https://maxcdn.bootstrapcdn.com/twitter-bootstrap/*` starting at `4.0.0-beta.*` please?

PR #874, opened 2017-10-19
Add Bootstrap 4.0.0-beta.2.

XhmikosR:
- [x] Update to the final beta2 release files
- [x] Verify SRI hashes
- [x] Make sure .map files and the new bundle files are on the CDN

@jdorfman @jmervine: please do NOT merge; I'll do it when it's ready. I open this to discuss the changes.


XhmikosR:
So, what should we do with the new bundle file? Also note that while we have reboot and grid only files on the CDN, we don't expose them currently. Not sure if we should.

/CC @mdo for thoughts too.

XhmikosR:
ping @jmervine @jdorfman: please add beta2 to `twitter-bootstrap` path too. That's why I had made #867...

jdorfman:
@XhmikosR restarting build

@jmervine can we just adjust the tests to ignore `https://maxcdn.bootstrapcdn.com/twitter-bootstrap/*` starting at `4.0.0-beta.*` please?

jdorfman:
@XhmikosR when was `twitter-bootstrap` uri changed to just `bootstrap`? We can stop checking versions after that, will speed up tests.

jdorfman:
@XhmikosR merge when ready

XhmikosR:
There's no point in skipping tests. You might as well remove the tests then...

The point is we need to remove `twitter` from the path somehow. Not sure when this started to break, probably in v4.

PR #875, opened 2017-10-19
Tweak ads style.

XhmikosR:


XhmikosR:
Maybe I should limit the text so that it's consistent with the image size? Something like `max-width: 400px;`?

XhmikosR:
I added a max-width so that the text has aprox. the same size as the image.

jdorfman:
@XhmikosR thank you for reminding me. We actually would like to move it to the top right i.e. above "Mad Love". Last month we raised ~$230 for the Open Source collective. We would like to see how it will perform above the fold. Easy?

XhmikosR:
Hmm, I'll have a go at it and let you know.

XhmikosR:
Admittedly, the current position is just too low in the page.

PR #880, opened 2017-11-01
Update report-uri URL

XhmikosR:
Now, this account is my personal one and with the recent changes, I'm not sure if we'll hit the limit or not. 

Maybe @ScottHelme can do something :)

PS. @ScottHelme maybe I missed something but couldn't the old report URLs be redirected?

### DIFF
--- a/app.js
+++ b/app.js
@@ -183,7 +183,7 @@ app.use(helmet.contentSecurityPolicy({
             'ghbtns.com'
         ],
         manifestSrc: ['\'self\''],
-        reportUri: 'https://d063bdf998559129f041de1efd2b41a5.report-uri.io/r/default/csp/enforce'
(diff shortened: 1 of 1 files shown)


ScottHelme:
The old URLs are being redirected but it's best to update your address. It saves a 301 and who knows what the future holds!

While you're here why don't you claim a vanity subdomain on the Setup page? They look a lot nicer than the random subdomain token:
https://report-uri.com/account/setup/

If you guys could clear a subscription with Team access it'd be awesome. You could break this out into a team and keep the reports separate with role based access (but they are still billed against the team owner). I might also be able to hit you up with some extra reports ;-)

XhmikosR:
The thing is that this is my personal account which I use for a couple of open-source sites, so I'm not sure what subdomain to set up so I left it at the default.

About the rest, it's up to @jdorfman :)

XhmikosR:
OK, I'm gonna merge this but I already exceed my quota.

PR #882, opened 2017-11-22
bootswatch4 beta

jmervine:
Addressing https://github.com/MaxCDN/bootstrap-cdn/issues/881

- Updating download scripts to support bootswatch 4.
- Adding new route.
- Adjusting beta placement in header (@jdorfman thoughts?)
- Adding bootswatch4 files.
- Fixing bug with bootswatch 3 images showing bootswatch 4 thumbnails.
- Fixing bug with bootswatch 3 links linking to bootswatch 4.


jmervine:
@jdorfman I have not uploaded the files to s3 yet. Wanted to get your feedback on this first.

docker-compose.yml
```diff
@@ -1,5 +1,5 @@
 web:
-  image: node:8
+  image: jmervine/herokudev-node:8
```
XhmikosR: What's this change about?

scripts/bootswatch.js
```diff
@@ -7,7 +7,9 @@ const path    = require('path');
 const fs      = require('fs');
 const request = require('request');
 
-const version = process.argv[2];
+const version  = process.argv[2];
+const verMajor = version[0];
```
XhmikosR: Can't we just do this programmatically instead of adding a new arg?

jmervine: It's not a new arg, it's taking the first character of the string.

docker-compose.yml
jmervine: Just using a different base image for my local development environment. One that more closely matches the deployment target. This doesn't impact the application running on heroku.

jmervine: If you run docker, your can use with the following command:

```
docker-compose run --rm --service-ports web bash

# then setup with
npm install

# and run with
node app.js

# or run tests with
node make test
# ... etc
```

scripts/bootswatch.js
XhmikosR: True, sorry, too tired :/

jmervine:
@jdorfman Cherry picked in your commit mentioned here:  https://github.com/MaxCDN/bootstrap-cdn/issues/881#issuecomment-346442293

PR #883, opened 2017-11-22
ignore twitter-bootstrap versions after 2.3.2

jmervine:
addressing https://github.com/MaxCDN/bootstrap-cdn/issues/877

### DIFF
--- a/tests/functional_test.js
+++ b/tests/functional_test.js
@@ -6,6 +6,7 @@ const assert    = require('assert');
 const walk      = require('fs-walk');
 const async     = require('async');
 const https     = require('https');
+const semver    = require('semver');
 const digest    = require(path.join(__dirname, '..', 'lib', 'helpers')).sri.digest;
 const helpers   = require(path.join(__dirname, 'test_helper'));
 const config    = helpers.config();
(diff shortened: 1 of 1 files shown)


XhmikosR:
Why don't you use \d? Also I'd skip adding a new dependency. The checks you
do above are fine in my opinion.

jdorfman:
1/3 faster than the previous test

jmervine:
@XhmikosR `\d` didn't work, given that it's just a test and I'm about to run out the door I didn't think it was worth the investigation. Also, assuming you mean `semver` by "new dependancy", it's already installed, just leveraging it.

XhmikosR:
There is no point in requesting any reviews from me any more since Justin
keeps merging stuff without me approving or rejecting.

jmervine:
@XhmikosR I find your reviews valuable, even if I don't take your suggestions. I know Justin was trying to get this out before I took off for the weekend and in conjunction with another PR, hence the rush. That said, even if merged and released, we can always address comments in a new PR.

XhmikosR:
The whole point of PRs, and any kind of code review, is to address issues
before something lands.

I've learned to work like that with anything and any other way just doesn't
make sense to me. It's just more work/time, you name it.

PR #886, opened 2017-11-25
functional_test.js: minor tweaks.

XhmikosR:
* use `string.includes()`
* use `\d` in the regex


jmervine:
Odd `\d` wasn't working for me, must have missed something. Didn't know about `String.includes` that's useful.

XhmikosR:
I wonder how it worked with period not being escaped to be honest.

BTW, if you want to keep the regex to a separate var let me know.

jmervine:
Not sure, I don't have a ton of experience with regex's in JS, although they feel odd compared to other languages, but I can't put my finger on exactly how.

I don't feel strongly about the regex being in a separate variable, I just borrowed that implementation from stack overflow as it was.

XhmikosR:
So, as long as the number of tests is the same, the regex works so we can merge.

PR #889, opened 2017-12-01
Switch to Bootstrap 4

XhmikosR:
**This is a WIP; do NOT merge until it's ready.**

- [x] Migrate all frontend code to v4
- [x] Make sure tests are passing
- [x] Make sure the breakpoint for showing the ads is the right one
- [x] Fix margin collapse
- [x] Show the bootstrap bundle since we already host it and use it?


XhmikosR:
@jmervine: can you help out with the test failures?

jmervine:
Yeah. I can take a look.

jmervine:
I don’t have a strong opinion, but my 2 cents is we leave the UI as is, only updating the underlying bootstrap framework.  Baby steps, you know?

The Try It buttons will need to be moved from the  bootswatch page to the bootswatch4 page. !85 otherwise the rest could remain the same.

XhmikosR:
Yeah but it's the only way Bootswatch will work otherwise it will be broken.

Feel free to push here

XhmikosR:
OK, tests fixed. Apparently the `sri` pug variable is needed in the pug files even though it's seemingly unused, otherwise tests fail...

jmervine:
Looks like the try it now buttons aren’t working.  At least not on my phone.

XhmikosR:
True. I'll fix that and a couple more issues I see on mobile.

XhmikosR:
OK, @jdorfman @jmervine: now this is ready for review; just confirm and I will merge when I believe it's 100% ready.

If you find any issues let me know.

XhmikosR:
I added the JS bundle in the frontend. We might want to tweak tests and stuff though.

PR #890, opened 2017-12-01
Legacy updates

XhmikosR:
**Please, do NOT merge!**

This PR needs to come after #889 and after we all agree with the layout changes.

@jmervine: not sure if this is the right way to do things; it seems to work but I had to specify the filename extention in `res.render`


XhmikosR:
So, for this, we need to decide the routes. Maybe we should use `/legacy/bootswatch/>version>` or something.

XhmikosR:
Oh, and I need help fixing the tests.

jmervine:
I would vote on holding off on this until bootstrap 4 is out of beta, personally. Additionally you'd simplify the implementation if you avoid the `legacy` sub-directory under opting for `bootstrap3` and `bootswatch3` instead.  

Regarding the failing tests, it looks like the bootswatch4 tests are filing because of the 301, the tests don't follow redirection. If you update [this line](https://github.com/MaxCDN/bootstrap-cdn/blob/develop/tests/bootswatch4_test.js#L7) to be `const uri = helpers.app(config, 'bootswatch');` it should fix that set of failures.  

The bootswatch tests are failing for a similar reason, https://github.com/MaxCDN/bootstrap-cdn/blob/develop/tests/bootswatch_test.js#L7 needs to be updated to match your changes.

XhmikosR:
@jmervine: feel free to push here please.

I don't think we should wait for v4 to hit stable. My only concern is us thinking the structure before we go through with this.

I'm still leaning to `/legacy/foo/version` instead of `/legacy/fooversion`.

jdorfman:
@jmervine I think we should stay ahead of the curve and start promoting BS4 even if it is in beta. getbootstrap.com is already promoting it as the default and we should be mirroring their interest. In fact, after this merge, I would like to swap `3.3.7` and `beta-2` position on `quick start`.

@XhmikosR besides the failing test you have an issue on `/bootswatch`

image

Let's get this out by Friday.

XhmikosR:
@jmervine: I redirected beta to root too and added the old v4 versions in legacy/bootstrap.

Maybe we should refactor our tests, so feel free to push.

XhmikosR:
So, we all agree with the current URLs? I'm not sure it's perfect, especially for Bootswatch. But I guess we can change it if needed later and redirect.

PR #892, opened 2017-12-04
Deploy to production

jdorfman:


XhmikosR:
I really don't like the fact that master is shown as behind. I know how to do it manually, but not sure how to do it via GitHub UI.

PR #896, opened 2017-12-05
refactoring config and specs

jmervine:
Creating this as a PR to #890 because I want to discuss before merging.

I ran out of time today before my meetings start, but in addition to these changes, I'd like to:

1. Rename "Legacy" to "Archives".
1. Remove the "Old Versions" dropdown, instead having "Bootswatch 3" and "Bootstrap Archives" nav buttons. My thinking is, "legacy/bootstrap" and "legacy/bootswatch" are actually different things. "legacy/bootstrap" is really more of an archive of older versions, which "legacy/bootswatch" is really "Bootswatch 3" (as I've renamed it in this PR).

### DIFF
--- a/views/index.pug
+++ b/views/index.pug
@@ -8,9 +8,10 @@ block content
     h3.text-center.mt-4=`v${config.bootstrap4[0].version}`
 
     .card.bg-light.mt-4.p-4.text-center
-        -var name   = 'quickstartCSS'.toLowerCase() + config.bootstrap4[0].version.replace(/\./g, '_');
-        -var file   = config.bootstrap4[0].stylesheet;
-        -var sri    = config.bootstrap4[0].stylesheetSri;
+        -var latest = config.bootstrap4[0];
+        -var name   = 'quickstartCSS'.toLowerCase() + latest.version.replace(/\./g, '_');
+        -var file   = latest.stylesheet;
+        -var sri    = latest.stylesheetSri;
         -var formId = name + '_form';
 
         .form-group.my-4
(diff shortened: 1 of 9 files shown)


jmervine:
Realized that the rename -- `tests/legacy_test.js → tests/bootstrap3_test.js` -- is probably inaccurate, given that it has both bootstrap 3 and 4.

jdorfman:
@XhmikosR review and merge when ready

jmervine:
@XhmikosR Squashed my commits up. Talked with @jdorfman about the suggestions I mentioned above and opted to leave as is. I'm hopping on a plane soon and won't be online for the rest of the day. Feel free to merge without me, otherwise I'll try and take a look at any comments this evening or tomorrow morning. o/

XhmikosR:
I will review it first thing, thanks. But you shouldn't have made a PR when you could just push to the same branch.

views/index.pug
```diff
@@ -8,9 +8,10 @@ block content
     h3.text-center.mt-4=`v${config.bootstrap4[0].version}`
 
     .card.bg-light.mt-4.p-4.text-center
-        -var name   = 'quickstartCSS'.toLowerCase() + config.bootstrap4[0].version.replace(/\./g, '_');
-        -var file   = config.bootstrap4[0].stylesheet;
-        -var sri    = config.bootstrap4[0].stylesheetSri;
+        -var latest = config.bootstrap4[0];
```
XhmikosR: Nice :)

jmervine:
Typically, I wouldn't make a PR, but in this case, I was changing somethings that I thought might warrant more discussion, as such, I wanted to have that discussion prior to pushing it to your branch.

PR #899, opened 2017-12-06
Replace our `helpers.contains` function with `String.includes`.

XhmikosR:


tests/integrations_test.js
```diff
@@ -22,32 +22,32 @@ describe('integrations', () => {
 
     it('contains authors', (done) => {
         config.authors.forEach((author) => {
-            helpers.assert.contains(author, response.body);
+            response.body.includes(author);
```
jmervine: There's no assertion around the check. I don't think is actually testing anything.

XhmikosR: True, sorry about that, missed it on copy paste. Do you want me to revert this whole patch or add assert?

jmervine: It's your call. The point of the helper was to wrap all that up in to a single call. The downside of using helpers in specs though is that it munges the stack trace a bit. Overall the test output on error would be a bit cleaner and easier to read if you add the assert with custom messages for each assertion. That's a lot of tedious and manual work though. Short term, I'd probably revert and update the helper method to use `includes`, if it doesn't already... but more for time sake, then it being the correct way to go.

PR #903, opened 2017-12-06
Fix and improve scripts/wp-plugin.js.

XhmikosR:


scripts/wp-plugin.js
```diff
@@ -7,32 +7,46 @@ const path   = require('path');
...
     bootstrap: {},
     fontawesome: {}
 };
 
-config.bootstrap.forEach((bootstrap) => {
+config.bootstrap4.forEach((bootstrap) => {
+    data.bootstrap[bootstrap.version] = {
+        css: bootstrap.stylesheet,
+        js: bootstrap.javascript,
+        jsBundle: bootstrap.javascriptBundle
```
XhmikosR: Only thing I don't know is if the WP plugin will have issues with this addition.

jdorfman: @petarpetrovic can you confirm?

jdorfman:
Holy on this until further notice.

XhmikosR:
Well, the script is broken currently without this PR.

I could remove the bs4 addition so that it works and merge for now, but my 2 questions still stand:

1. Why do we need the script since we generate data/bootstrapcdn.json via express?
2. Will the WP plugin throw with the `jsBundle` entry?

XhmikosR:
@jmervine: are we sure the bootstrap4 addition will not break the WP plugin?

Also, why do we need the script since we generate the json file server-side?

jmervine:
No, I'm assuming it will break the WP plugin basedon @jdorfman's comments. 

I approve the code change though, it all looks good. I'd assume you'd wait for the go-ahead from Justin to merge.

XhmikosR:
All right, but do we need the script anymore?

jmervine:
Oh, I see what you're asking now. Not 100% sure, but I doubt it unless someone is uploading the result to S3.

XhmikosR:
Should I just remove it then and all its references?

XhmikosR:
I removed the bootstrap 4 addition.

I added one more patch to remove the script and its references, so just let me know which way we should go.

XhmikosR:
I'm merging this as is, without any removal, but we should revisit it and remove it if it's unused.

PR #907, opened 2017-12-06
fixing data/bootstrapcdn.json

jmervine:


XhmikosR:
Maybe add Bootstrap 4 too above bootstrap 3?

Also, if we no longer need scripts/wp-plugin.js we should remove all references/files.

jdorfman:
@XhmikosR not yet. @petarpetrovic needs to update the plugin https://cl.ly/252L232n3l0G

XhmikosR:
I don't think what you pasted above is relevant...

PR #908, opened 2017-12-07
Load Twitter timeline manually.

XhmikosR:
This allows us to skip loading it when we are on < 992px, where we don't show the timeline anyway.

Should make mobile browsing much faster, we need to test this good before merging.

`window.matchMedia` is available for our supported browsers so we should be all good there.

### DIFF
--- a/public/assets/js/main.js
+++ b/public/assets/js/main.js
@@ -71,6 +71,25 @@
 
     })();
 
+    function initTwitterTimeline() {
+        if (!window.matchMedia('(min-width: 992px)').matches) {
+            return;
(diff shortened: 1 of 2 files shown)


XhmikosR:
Before: https://www.webpagetest.org/result/171207_ZB_3d01afe1b58ac7cc89f290385c19ad5e/1/details/
After: https://www.webpagetest.org/result/171207_D7_6ca79b01b728be48ec26063f035eb35c/1/details/

jdorfman:
Great work @XhmikosR

PR #910, opened 2017-12-07
Run HTML validator for all pages.

XhmikosR:
Small test slowdown, but we check everything this way.


jdorfman:
@XhmikosR merge when ready

XhmikosR:
I'm down with the flu, so better someone else merged this and deployed.

PR #912, opened 2017-12-11
Removing assert.contains helper for better failure messaging

jmervine:
1. Removed `helper.assert.contains` in favor of `assert(response.body.includes(...), "messaging")`. Currently find debugging test failure a bit annoying due to lack of proper messaging.
2. Removed `helper.assert.response` in favor of in-test asserts. With the helper, the failure messaging was obfuscated. 

I chose to leave the other assert helpers for now as they provide a bit more "help".

### DIFF
--- a/tests/showcase_test.js
+++ b/tests/showcase_test.js
@@ -1,6 +1,7 @@
 'use strict';
 
 const path     = require('path');
+const assert   = require('assert');
 const helpers  = require(path.join(__dirname, 'test_helper.js'));
 const config   = helpers.config();
 const uri      = helpers.app(config, 'showcase');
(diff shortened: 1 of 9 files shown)


tests/showcase_test.js
```diff
@@ -26,32 +28,37 @@ describe('showcase', () => {
 
     it('contains authors', (done) => {
         config.authors.forEach((author) => {
-            helpers.assert.contains(author, response.body);
+            assert(response.body.includes(author), `Expects response bodt to include "${author}"`);
```
XhmikosR: Small typo `bodt`.

XhmikosR:
Looks good as long as tests pass, but couldn't we just pass an extra param with the message to our `helper.assert.response`? Not sure which one is better, it's your call :)

jmervine:
Passing an extra param would have worked as well, but at the point, there's really no difference between the two and this feels clearer to me. Especially for someone coming in that's new to the project.

XhmikosR:
Yeah, not sure it would have been better either.

PR #913, opened 2017-12-11
adding fontawesome legacy page

### LINKED ISSUE #911: Add Font Awesome to Old Versions
Someone pinged me and asked if we could list the older versions of Font Awesome just like we do with the others.

jmervine:
Fixes #911


jmervine:
@jdorfman Adding this, but my concern with shipping it is that I'd prefer not to do this with bootswatch -- the UX would be annoying at best. This sets a precedent that could lead to that request as well.

XhmikosR:
Yeah, definitely not with Bootswatch. We'd need another implementation for that.

This looks good as far as I can tell.

XhmikosR:
Hmm, tests broke due to the helpers refactoring.

@jdorfman; please enable the option for the branch to always be on sync with develop otherwise prevent merging.

PR #921, opened 2017-12-12
Clean up `require`s.

XhmikosR:


XhmikosR:
@jmervine: any objections to merge this?

jmervine:
What include the `.js`, it seems to be unnecessary cruft. Is there a benefit aside from perhaps distinguishing from what's local or not?

XhmikosR:
That but also distinguish if it's a folder or not. For example for routes/index.js, we can just do `require('routes')`.

See https://nodejs.org/api/modules.html#modules_file_modules

PR #924, opened 2017-12-12
Fix multiple Twitter timelines under some cases.

XhmikosR:
So, if one was on >= 992px and resized the window, they ended up with multiple timelines.


public/assets/js/.eslintrc
```diff
@@ -7,3 +7,4 @@ rules:
     no-var: off
     object-shorthand: off
     prefer-arrow-callback: off
+    prefer-template: off
```
jmervine: Is there a way to disable this inline instead of globally? I prefer the `+` concatenation employed below in instances like that, but in general, would like eslint to bark in favor of templates.

XhmikosR: Templates are used where we use es6, i.e. not in public/js. This file is specific for the files in this folder.

PR #925, opened 2017-12-12
Move footer outside of the main content.

XhmikosR:
* Move ads to the left of the main content, keep Twitter timeline to the right
* Move footer outside of the main content

This is a first idea. I've always been meaning to move footer outside of content.

Let me know if you have any suggestions to improve this.

### DIFF
--- a/views/_partials/footer.pug
+++ b/views/_partials/footer.pug
@@ -1,25 +1,28 @@
 footer.mt-5.py-4.text-center
-    p
-        | Maintained by 
-        a(href='https://twitter.com/jdorfman', target='_blank', rel='noopener') @jdorfman
-        |  , 
-        a(href='https://twitter.com/mervinej', target='_blank', rel='noopener') @mervinej
-        |  and 
-        a(href='https://xhmikosr.io/', target='_blank', rel='noopener') @XhmikosR
-    p
-        a(href='https://www.stackpath.com/', target='_blank', rel='noopener')
-            include /assets/img/stackpath-icon.svg
+    .container
+        .row
+            .col-12
+                p
+                    | Maintained by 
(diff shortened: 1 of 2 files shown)


jmervine:
I'm happy with the footer, but personally, prefer the ads where they were. I feel like this squeezes the center column too much on smaller displays.

XhmikosR:
How small you mean? Because currently I have it use the full row up until md breakpoint.

jmervine:
Here's what I'm looking. Browser width is currently 1650px.

screen shot 2017-12-12 at 9 19 01 am

jmervine:
Perhaps removing padding from the left and right might help?

XhmikosR:
Well, that is normal... Because the biggest breakpoint in >= 1200px, and I use 2 + 5 + 2 columns, otherwise on very big resolutions the content is full width. There's no padding, just flexbox.

jmervine:
Fair. I still prefer the ads where they were, on the right. But I don't feel super strongly about it.

I'll leave it to @jdorfman to decide.

XhmikosR:
Have one more look, I made everything use a little more space.

jmervine:
It looks better, for sure. I still prefer it the way it was, on the right, but I'll defer to @jdorfman.

XhmikosR:
All right, let me know, and I can revert the ads move and just keep the footer change.

XhmikosR:
So, I reverted the ads move. I think I can merge now.

jdorfman:
@XhmikosR the deploy link is dead

XhmikosR:
@jdorfman: not related to this, something in the integration must fail. @jmervine: ideas?

PR #928, opened 2017-12-12
Stop loading Twitter and GitHub buttons `onload`.

XhmikosR:
It seems it's actually slowing us down.

Testing this for now.

### DIFF
--- a/public/assets/css/style.css
+++ b/public/assets/css/style.css
@@ -21,11 +21,6 @@ iframe {
     overflow: hidden;
 }
 
-#ghbtns-badge {
-    width: 110px;
-    height: 20px;
-}
-
 footer {
     border-top: 1px solid rgba(0, 0, 0, 0.1);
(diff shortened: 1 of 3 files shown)


jdorfman:
seems slower than before ¯\_(ツ)_/¯

jmervine:
You can't compare a PR App to production. They auto-idle after 5 minutes, and have to wake up on first hit. As such they'll always be a bit of added latency when you first hit it. Additionally, the "hardware" there has a fraction of what's on production. If you want a side by side you need to cut a branch that's identical to develop and create a PR for that. Hit them both once to ensure they're not idling and then run a web page speed test on them.

XhmikosR:
It seems a middle solution is to load the third-party script on `DOMContentLoaded`.

jdorfman:
lgtm. New rule if you open it and get approval you merge it.

PR #929, opened 2017-12-14
Use LoadJS to load 3rd-party JS asynchronously.

XhmikosR:
The pros:

1. We no longer have any block-rendering JS leading to better load times (more than 500ms decrease in first meaningful paint, similar decrease in other timings)

The cons:

1. We can't use integrity for the files we load via LoadJS, as far as I can tell so far.
2. Less important; I couldn't make Pug work with an include for our jQuery code inside `script.` so I had to hardcode it in `layout.pug`. Our current jQuery code is this:

    ```js
    $(function () {
        $('.ads-info-toggler').popover();
    });
    ```

EDIT: Second issue fixed.


views/layout.pug
```diff
@@ -77,19 +77,25 @@ html(lang='en', prefix='og: http://ogp.me/ns#', itemscope, itemtype='http://sche
...
-            include /assets/js/clipboard.min.js
+            include /assets/js/vendor/clipboard.min.js
+            include /assets/js/vendor/loadjs.min.js
 
             if (process.env.NODE_ENV === 'production')
                 include:uglify-js /assets/js/main.js
             else
                 include /assets/js/main.js
 
+            | loadjs(['#{config.javascript.jquery.uri}', '#{bootstrap.javascriptBundle}'], 'jquery', {
```
jmervine: If you can't move all of this in to the include, per your comment, I would just move the contents of the include in to the pug template as well. It will read better.

XhmikosR: I don't think that's possible because of the variables. If you make it work, let me know.

XhmikosR: I mean, move it to a separate JS file. A pug file should work fine, though.

views/layout.pug
```diff
@@ -77,19 +77,25 @@ html(lang='en', prefix='og: http://ogp.me/ns#', itemscope, itemtype='http://sche
...
+            |     async: false,
+            |     numRetries: 3
+            | });
+            | loadjs.ready('jquery', {
+            |     success: function() {
```
jmervine: Could probably just make this...

```javascript
loadjs.ready('jquery', {
    success: $('.ads-info-toggler').popover
});
```

XhmikosR: True, changed it, and move the whole block to a pug include. The jQuery code is still hardcoded, not sure if it's worth having it separate since it's just one line.

jmervine: Yeah, that's what I was thinking. Not really worth it.

jmervine:
I like the second approach a lot more! :D

PR #930, opened 2017-12-15
Switch to xo

XhmikosR:
So, this is opinionated too, but I believe it's better to use such tested config.

Only downside so far, is that it's using ESLint 3.x, but I guess they'll update it eventually.

BTW, I tried to match our current style as close as possible.


XhmikosR:
I reverted the filename changes and ignored those, I don't have a strong opinion either. :)

XhmikosR:
I'm gonna split the changes and merge separately without switching to xo, at least not yet.

PR #937, opened 2017-12-20
Improve cache busting

XhmikosR:
@jmervine: here's the pros and cons

1. My solution works fine too, but it's like reinventing the wheel when there's a package that does the same. Also the staticify package provides a middleware so we don't need to redirect the hashed assets ourselves. **With my solution the hashed assets are served with a cache of 0 so we'll need to tweak this if we go with it.**
2. staticify's middleware has its own settings, so what we set for serve-static, doesn't apply here. Not a huge deal since we can provide upstream patches and improve it, but currently the hashed assets are served with **a cache of 1 year (which is OK) but with an etag set**.
3. staticify causes a slowdown when the server process starts, because it scans the whole directory. We could provide an upstream patch to improve this.
4. The advantage of all this is that our hashed assets now, will still be in the browser's cache even when our server process is restarted, since their hash is independent of the date the server process was started.

[Here](https://github.com/errorception/staticify/issues) are the issues I have currently opened, any help to improve this for our needs (and everyone else's) appreciated :)

Let me know with which solution we should go. I prefer the staticify solution and provide upstream patches.

### DIFF
--- a/app.js
+++ b/app.js
@@ -83,12 +83,6 @@ app.use(compression());
 
 app.use(favicon(path.join(PUBLIC_DIR, config.favicon.uri), '7d'));
 
-app.use(serveStatic(PUBLIC_DIR, {
-    maxAge: '30d',
-    lastModified: true,
-    etag: false
-}));
-
 app.use((req, res, next) => {
     // make config available in routes
     req.config = config;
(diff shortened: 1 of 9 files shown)


jmervine:
Dig. Do you want to wait for the upstream patches to be merged before moving forward with this, or is it acceptable as it, but improved with the patches?

XhmikosR:
I'm not sure either... I mean, with my solution we don't have the proper
max age so it beats the point unless we set it explicitly... which
currently I'm not sure how to set the max age for the redirected assets.

The other solution needs some more work.

So if we could fix the expires date we could switch to my solution and
switch to staticify when it's updated.

PS. Have you tried any caching plugins for express ? Or using cluster?

jmervine:
I haven't tried any of the express plugins. At least not that I remember.

XhmikosR:
I think you used cluster in the past for this repo. Do you remember why you
removed it?

jmervine:
By cluster, I'm assuming you mean https://nodejs.org/api/cluster.html

I removed it for two reasons. First cluster's load generating randomization at the time was spotty at best. It would occasionally over load a process, directing too much traffic to one. I'd assume -- hope -- they've fixed that by now. The second is that it was added when we were launching the app on a single (non-Heroku) host. On Heroku, it makes more sense to run smaller dynos (instances) and launch more of them. Running multiple processes isn't really the best architecture, unless you have really high load.

PR #938, opened 2017-12-20
Remove Glyphicons from the Resources dropdown.

XhmikosR:
Bootstrap 4 doesn't use it.


XhmikosR:
Well, it might be too soon, I'd give you that. But other than that, it doesn't make sense for us to keep having it since it's no longer used. I mean, what if a component was used in v2.x, would we still list it?

jmervine:
@jdorfman your call

jdorfman:
@XhmikosR merge when ready. WordPress should be next. If @petarpetrovic ever get around to updating it we can re-add it.

XhmikosR:
I removed WordPress plugin too.

PR #941, opened 2017-12-22
Production release

XhmikosR:


views/bootlint.pug
```diff
@@ -14,7 +14,7 @@ block content
             .input-group.input-group-lg
                 input.form-control(id=formId, type='text', value=file)
                 span.input-group-btn
-                    button.btn.btn-secondary(type='button', data-toggle='collapse', data-target='#' + name, aria-expanded='false', aria-controls=name)
+                    button.btn.btn-dark(type='button', data-toggle='collapse', data-target='#' + name, aria-expanded='false', aria-controls=name)
```
jdorfman: @XhmikosR what is the diff? I don't see anything visually.

views/showcase.pug
```diff
@@ -4,15 +4,15 @@ block content
     h2.text-center.mb-4 Showcase
 
     .row
-        .col-md-6
+        .col-12.col-md-6
             each item in col1
                 include _partials/showcase.pug
-        .col-md-6
+        .col-12.col-md-6
```
jdorfman: @XhmikosR what's the diff besides the class name?

jdorfman:
@XhmikosR merge when ever

views/bootlint.pug
XhmikosR: Well, I shouldn't have used `secondary` in the first place, since we wanted dark. It just happened to look the same for the theme we were using.

views/showcase.pug
XhmikosR: No difference; one could drop `.col-12` too, but I prefer being explicit.

XhmikosR:
@jdorfman @jmervine: I need to go, so if everything looks good to you, approve and merge. The old approval is lifted after something else is pushed.

XhmikosR:
@jdorfman: you might want to disable this feature regarding reviews if something is pushed. Your call.

jmervine:
This LGTM, I'm thinking we should get https://github.com/MaxCDN/bootstrap-cdn/pull/942 in to it as well though.

XhmikosR:
If you think it's fine sure go ahead. I'm afk. But I still don't get why
the max age isn't being set. Not a blocker but you know.

PR #947, opened 2018-01-03
Add Privacy Policy page.

### LINKED ISSUE #864: Add privacy policy page
/CC @jdorfman

XhmikosR:
Fixes #864 

- [x] Add tests
- [x] Update sri.js to generate the hashes for any css/js


XhmikosR:
@jdorfman: is there any way we can just get the HTML, CSS and JS and we load it from our server? We will keep their button and everything, I just want to simplify our CSP and make sure the page is fast and works even without JS.

jdorfman:
@XhmikosR just extract and convert the HTML to pug. I provided you with everything they give us. 

BTW this looks horrible: 
image

It should go under `Homies be`

XhmikosR:
The branch isn't ready, I'm aware of the issues.

XhmikosR:
All right, it should be sort of ready now. It's so much bloat but it seems to work. The rest of the pages shouldn't be affected.

@jmervine: if you could add some tests that would help. Or help with the integrity for the new css file.

XhmikosR:
Also, I'm seeing numerous Application Errors on the heroku instance.

```
 Application error

An error occurred in the application and your page could not be served. If you are the application owner, check your logs for details. 
```

jdorfman:
@XhmikosR looks great! Merge when ready. As far as the Heroku errors go, @jmervine might know.

XhmikosR:
We need tests though. And I need to remove the double header... and maybe
update the integrity script to automatically generate the SRI hash for the
new file.

XhmikosR:
Also note to self to remove the part that says they collect data since we
host everything and it doesn't apply anymore

XhmikosR:
I need to confirm one last thing regarding collapsed sections, so please do not merge yet.

Issue #948, opened 2018-01-03
Change bootswatch.js script to use GitHub tags


XhmikosR:
Currently we get the files from the live server. This is bad, because we don't follow any version control. So if one runs `npm run bootswatch 3.3.7` you will see the files will differ.

We should retrieve the GitHub tag instead.

@jmervine: if you make any progress just ping me.

jdorfman:
@jmervine @XhmikosR what is happening with Bootswatch?

People are asking: https://twitter.com/orgbrat/status/968516215634321410

XhmikosR:
@jdorfman: First of all, it's unrelated to this issue.

Secondly, it's in develop, not in master yet.

PR #951, opened 2018-01-05
Add bootswatch 4.0.0-beta.3.

XhmikosR:
This is untested since the files aren't on the CDN yet.


XhmikosR:
ping @jdorfman for adding the files on the CDN

jdorfman:
Uploaded, waiting for test to pass

jdorfman:
@jmervine I would like to remove checking "bootstrap.css" in Bootswatch going forward. No one should be using anything other than `.min.`

Willing to talk about it, but rather just have it changed.

jmervine:
We check in the non-minified versions to allow people to view and deconstruct the contents with out having to un-minify the min version. It’s probably not a very common use-case though. 

Your call really. I don’t have strong opinions.

XhmikosR:
Well, we host many files we don't test. That being said, it's always better to err on the safe side.

If we feel tests are getting too slow again, maybe we should randomly test 10 files or something. It should be better than not testing anything.

Now, about this PR. I restarted the Travis tests. If they are green, I'm gonna merge it.

XhmikosR:
OK, tests are still failing. @jdorfman: Not sure where you got the ones you uploaded but the hashes are wrong. The files I have committed in this branch are from the upstream git tag.

jdorfman:
@jmervine `.min` looking at the logs, bootswatch is low as it is.

@XhmikosR weird I downloaded them directly from their website. I'll check it after work.

XhmikosR:
@jdorfman: you mean you downloaded each css file separately? Either way, I vote we go with the git tag and I also have an issue to improve our script for this (see #948).

Also, @thomaspark, if you make changes to the live files, can you keep a consistent flow, at least from now on? Like only upload tagged releases or something.

jdorfman:
@XhmikosR ah! I didn't even think to check https://github.com/thomaspark/bootswatch/releases

I wrote a bash script a while a go that downloaded the css files. I'll re-upload those.

jdorfman:
@XhmikosR from now on I will just `git pull` and not use the script since it is breaking the file integrity.

```
> sri384 https://bootswatch.com/4/cerulean/bootstrap.min.css
sha384-SJRp3lIuvy82g2roYNoSMWGEf7ufogthbl9noRHbJIxsMPcFl6NEi5fO3PuYptFv
 ~ 
> sri384 https://maxcdn.bootstrapcdn.com/bootswatch/4.0.0-beta.3/cerulean/bootstrap.min.css
sha384-SJRp3lIuvy82g2roYNoSMWGEf7ufogthbl9noRHbJIxsMPcFl6NEi5fO3PuYptFv
```

XhmikosR:
Yeah, we definitely need to fix #948 :/

PR #955, opened 2018-01-07
Limit tweets to 4.

### LINKED ISSUE #953: Clean up tweets / reduce the number of visible ones
1. I'd say just remove the simple ones which are like "Thank you"
2. See if there's a way to reduce the number of visible tweets

XhmikosR:
~~**Do not merge yet.**~~

If we specify a limit, then the height is ignored. We don't need to change to a collection, but if you guys prefer it, we change it.

Fixes #953

### DIFF
--- a/public/assets/js/main.js
+++ b/public/assets/js/main.js
@@ -79,8 +79,8 @@
                 document.querySelector(timelineSelector),
                 {
                     dnt: true,
-                    height: 525,
-                    partner: 'tweetdeck'
+                    partner: 'tweetdeck',
+                    tweetLimit: 4
                 }
             );
         });


jdorfman:
Make it 4. I don't like it being longer than the `v3.3.7 card`

image

XhmikosR:
But do we like this approach?

How about you just remove some tweets from the timeline and we don't limit
them at all? The only downside is that when you add tweets you should
remove some others so that we keep it to a small number.

jdorfman:
@XhmikosR I don't just like the approach, I love it. This is how we will do it.

jdorfman:
@XhmikosR update to 4 please

XhmikosR:
Haven't had the time to work on this, but I'll get to it in soon(™)

XhmikosR:
@jdorfman: done.

jdorfman:
@XhmikosR perfect. MWR (merge when ready)

Did this make a significant improvement to webperf? I would like a before an after.

XhmikosR:
I'll make a test when we deploy to master and let you know.

XhmikosR:
@jdorfman: https://gtmetrix.com/compare/IiGZOFWz/wB4Txwpq

This is the live site with the 4 tweets vs after you deleted those tweets.

PR #965, opened 2018-01-18
adding bootstrap 4.0.0 \o/

jmervine:
I grabbed the files from s3, so they should match perfectly.


jdorfman:
@jmervine merge when ready

PR #973, opened 2018-01-22
Move JS outside of HTML and preprocess it.

### LINKED ISSUE #961: Move JS outside of HTML
I need to test the performance impact, because, well, we don't have http2 on Heroku.

But, this way, the assets will be cacheable, so it should save us bandwidth.

Note to self: make sure SRI hashes work properly when we minify the files on the fly. Otherwise if we can't get this to work, we might 1) either drop minification on the fly, 2) consider preprocessing the assets and just serve them.

XhmikosR:
Async load it, but still inline main.js.

Fixes #961.

@jmervine: can you think of a way to use babel and uglify without inlining main.js?

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,6 +1,7 @@
 /.git/
 /node_modules/
 /public/assets/js/vendor/
+/public/assets/js/main.{dev,prod}.js
 /public/bootlint/
 /public/bootswatch/
 /public/html5shiv/
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,6 +1,7 @@
 /.git/
 /node_modules/
 /public/assets/js/vendor/
+/public/assets/js/bundle.{dev,prod}.js
 /public/assets/js/main.{dev,prod}.js
 /public/bootlint/
 /public/bootswatch/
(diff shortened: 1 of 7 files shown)


jmervine:
@XhmikosR I've never used babel and haven't used uglify in years, so I don't really have much to contribute without some digging.

XhmikosR:
It's a general question, though.

I want to move our main.js file and load it asynchronously. But it needs to be passed though Babel and uglify-js. The latter isn't mandatory; we can live with such a small snippet being non minified. But it has to be parsed through Babel.

XhmikosR:
I think we need to preprocess the JS (and CSS if we are going this way). I can't find any way to do this on the file with separate files.

But since we are going to load everything asynchronously, I don't think there will be any performance descrease. Well, apart from not using http/2.

Not sure how to proceed... inlining everything (what we currently do) is "easier", but we don't have the ability to cache the JS files plus the HTML size is bigger than what it could be, resulting in slower initial time.

XhmikosR:
I switched to preprocessing JS files. @jmervine please have a look when you have time.

XhmikosR:
Also, note that I don't commit the generated files to the repo. We could do that if needed, but currently I run the scripts before the server runs.

PR #909, opened 2017-12-07
Add Font Awesome 5

jdorfman:
New structure. Instead of CSS they are now using .js

So we would have to do: `5.0.0/svg-with-js/js/fontawesome-all.js`

E.g.: this is their [Getting Started](https://fontawesome.com/get-started) code:
`<script defer src="https://use.fontawesome.com/releases/v5.0.0/js/all.js"></script>`

PR #975, opened 2018-02-10
Create v5.0.6 (new version of Font Awesome)

saxobroko:


jmervine:
@saxobroko Thanks for the contribution, but we're not currently supporting Font Awesome 5.x (see the thread here https://github.com/MaxCDN/bootstrapcdn/pull/909). Additionally, we were to consider merging a PR, the files would have to be added in the same format as the others (see version 4.7.0 for an example https://github.com/MaxCDN/bootstrapcdn/tree/develop/public/font-awesome/4.7.0) and [the configuration](https://github.com/MaxCDN/bootstrapcdn/blob/develop/config/_config.yml), app routing and UI would need to be updated to support FA 5.x's new formatting.

Feel free to weigh on the discussion on https://github.com/MaxCDN/bootstrapcdn/pull/909 regarding FontAwesome 5.x.

Cheers!

PR #971, opened 2018-01-22
Add Bootswatch v4.0.0.

XhmikosR:
@jdorfman: please upload the files on the CDN and make sure the `latest` symlink works fine.

Fixes #968

PR #976, opened 2018-02-10
Fixing symlinks for tests.

jmervine:
@jdorfman @XhmikosR assuming the specs on this pass, you should be able to merge it in to your branch -- https://github.com/MaxCDN/bootstrapcdn/pull/971 -- and be good to ship that.


XhmikosR:
But this is intentional. We don't want to bump the latest. Instead you should revert ASAP the CDN files to the ones from the symlinks.

jmervine:
I didn’t change the S3 at any point. 

I can revert the latest symlink, but the config will need to be changed to reflect latest as which ever version is symlinked. That will also display as the primary version on the site as well. 


This makes no sense though. Why is 4 not the latest? That breaks the pattern we’ve always used.

jmervine:
@jdorfman ^^?

XhmikosR:
I don't see why the config needs to be changed. It doesn't affect the CDN and it's only for the frontend code.

So, yes, please sync the files like they are in the develop branch.

XhmikosR:
As to why, it makes **total** sense. v4 is a major version and it **will** break every dependent site. The `latest` feature should have never been implemented like that. It needs proper version handling (like `latest@3` etc) and I for one can't do that with S3.

jmervine:
@jdorfman I’ll leave this to you. Latest has never been advertised AFAIK and we’ve never treated it as such. The expectation had always been that latest would break on update. Perhaps pull numbers on its usage?

If anything, we should just remove latest all together. IMHO.

XhmikosR:
This is something I've already discussed with Justin. It's not my own decision ;)

XhmikosR:
As for its removal, consider it deprecated. And we remove it after a few months or something like that.

jdorfman:
I think `3.3.7` is a good version to stop at with `latest`. We won't remove the symlinks, or update them. They will just be stuck in time. `4.0.0` is a clean slate.

jdorfman:
@jmervine let's just remove `/latest` from the tests. No longer supported.

PR #977, opened 2018-02-10
mirror s3

jmervine:
@jdorfman thoughts? close if no.


jdorfman:
I think all we need to do is just enable it in the AWS console. Every time I try, it fails. Not sure if a screen share would be good or not. I'll leave this open until we can address it.

jmervine:
@jdorfman All this PR is doing is making github match what's currently in s3. Including file permissions. This is the result of running an s3 to local sync. They should match and these files should be `0644`, basic rw for files, as opposed to `0755`, which is rwx for directories and executables.

XhmikosR:
The permissions change is fine. The image changes too; my guess is I optimized it some time in the past when I shouldn't have touched those.

But, the new bootstrap v4 files are redundant. We never exposed them...

XhmikosR:
Also, v3.3.7 doesn't have a bundle file; I have no idea where you guys took it from...

Finally, having the old jquery in the CDN seems bad to me. We didn't expose it ever to the public.

jmervine:
@jdorfman @XhmikosR  -- Now that the tests are passing, I'm going to merge this unless there are any objections.

jdorfman:
go for it

jmervine:
No rush on deploying this, it can go whenever.

XhmikosR:
Just don't forget to remove the removed files from S3 too so that we are in total sync.

Issue #949, opened 2018-01-03
Serve CDN files with charset=UTF-8


XhmikosR:
Not sure why we don't do this already...

We should serve any supported file types with a specified charset.

/CC @jdorfman

jdorfman:
@petarpetrovic any ideas?

petarpetrovic:
@XhmikosR can you give me an example file, both the origin URL and CDN URL?

cc @jdorfman

jdorfman:
@petarpetrovic for intance:

```
> icurl https://cdn.jsdelivr.net/npm/jquery@3.2
content-type: application/javascript; charset=utf-8
```

```
> icurl https://maxcdn.bootstrapcdn.com/bootstrap/4.0.0-beta.3/js/bootstrap.min.js
content-type: application/javascript
```

I am sure we can set an EdgeRule to add it, but I think it is something that needs/should be enabled on the origin side.

jmervine:
@jdorfman I could write a script crawl either [`config/_config.yml`](https://github.com/MaxCDN/bootstrap-cdn/tree/develop/config/_config.yml) or [`public`](https://github.com/MaxCDN/bootstrap-cdn/tree/develop/public) and adjust the content type of each element in s3. I don't love that approach though. It feels, risky.

Additionally, we could look at something like this: https://serverfault.com/questions/725562/recursively-changing-the-content-type-for-files-of-a-given-extension-on-amazon-s

XhmikosR:
So we need to specify the file types ourselves ? If so I say we go with a
well know list of files like the one from html5 no apache config.

jdorfman:
@jmervine I agree with you. It does seem risky for very little reward (IMO).

With that said, we have a few other S3 tasks:

 - `*.bootstrapcdn.com/index.html` -> redirect -> bootstrapcdn.com
 - EU replication

XhmikosR:
Oh and add tests so that we don't hit this again without us noticing :)

XhmikosR:
What do you mean little reward? This is wrong as any Unicode characters
won't be properly displayed without the right charset...

jdorfman:
@XhmikosR without Googling, what is the advantage of adding `; charset=utf-8`?

XhmikosR:
What I say in my previous message. This isn't an enhancement. It's a bug.

jmervine:
[...]

PR #978, opened 2018-02-10
Checking js and css content types more exactly.

jmervine:
@jdorfman This will provide better testing for updated contents once #949 is complete.

### DIFF
--- a/tests/functional_test.js
+++ b/tests/functional_test.js
@@ -69,7 +69,7 @@ function assertSRI(uri, sri, done) {
 
 const s3include = ['content-type'];
 
-function assertHeader(uri, header) {
+function assertHeader(uri, header, value) {
     if (typeof process.env.TEST_S3 !== 'undefined' && !s3include.includes(header)) {
         it.skip(`has ${header}`);
     } else {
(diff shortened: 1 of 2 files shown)


jmervine:
PS it will fail terribly before hand. =D

jmervine:
Once the cdn is purged, kick off these Travis runs again and I believe they’ll pass. I’ll f they don’t I’ll check it out. Just let me know once they’ve run. 

This PR is tests only so it can be merged once passing.

jdorfman:
@jmervine I want to wait for @petarpetrovic or someone on the StackPath side to monitor the logs right after I purge all just in case. Cool?

XhmikosR:
I restarted Travis but tests still seem to fail.

XhmikosR:
It seems that map files need to be re-uploaded with the new content type.

Anyway, after this is solved we can merge #971 .

jmervine:
This has been run on `public/js` and `public/css` which was missing before. Additionally I reran it for all `map` files. 

@jdorfman The CDN needs a purge for `*.map` files as well as the two directories mentioned above.

jdorfman:
@jmervine sorry just saw this =/

Purged and restarted the test

PR #983, opened 2018-03-13
Update dependencies.

XhmikosR:


XhmikosR:
OK, I fixed the pug errors; the filters are applied in reverse order, thus uglify was being fed with the ES6 code and choked.

XhmikosR:
Any ideas why tests time out?

PR #985, opened 2018-03-16
Asynchronously load all CSS and JS.

XhmikosR:
CSS is asynchronously loaded only for our main theme.

The critical.css needs to be manually generated when the CSS changes.

----

**I'm opening this so that you guys see the performance improvement in action.**

Now the content is shown almost instantly.

In order to use this we need to automate the critical.css generation as a preprocess step.

Unless there's some express.js plugin to do this, which I couldn't find one...


jdorfman:
@XhmikosR am I missing something? The _**perceived**_ performance boost isn't noticeable to me.

XhmikosR:
Well, you need to visit the link at least one time and the second time you can see it; just use an incognito tab the second time. It's because of how the Heroku PR integration is set up.

XhmikosR:
Here's a before and after comparison using Lighthouse

before &bull; after

XhmikosR:
After we sort out #973, I should be able to automate critical.css generation.

jdorfman:
@XhmikosR That is awesome. Amazing work.

jmervine:
I think the failing specs were a blip. The first one I restarted succeeded so I'm restarting the other as well.

XhmikosR:
So... it seems that the file that is not committed in git 404s.

XhmikosR:
@jmervine: what command does Heroku run? If it's `npm start` this shouldn't happen.

XhmikosR:
OK so we need to solve the following:

1. What is the exact command Heroku runs (dev/prod/PRs) and the value of `NODE_ENV`
2. If Heroku runs the npm scripts we shouldn't need to commit the generated bundle and critical css files
3. For some unknown reason AppVeyor fails while it works fine on my local Windows dev machine... maybe the `PORT` variable isn't being respected there

jmervine:
Heroku doesn’t run anything we don’t tell it to run. This is what we tell it to run 

https://github.com/MaxCDN/bootstrapcdn/blob/develop/Procfile

As for NODE_ENV I’ll double check when I get to a computer but it should be set to “production”. Again that’s something we set.

XhmikosR:
I know it doesn't but I didn't know how you had set it up exactly. :)

So, in this case I will change the Procfile to do npm start, in a little
while.

package.json
```diff
@@ -17,14 +18,14 @@
     "mocha:functional": "mocha --timeout 15000 --slow 500 --exit -R tap ./tests/functional_test.js",
     "dev": "npm run watch",
     "watch": "npm-run-all --parallel watch:*",
-    "watch:js": "cross-env NODE_ENV=development nodemon --ext js --ignore tests --ignore public/assets/js/main.dev.js --ignore public/assets/js/bundle.dev.js --exec \"npm run build:development\"",
-    "watch:other": "cross-env NODE_ENV=development nodemon app.js --ext css,pug,svg,png,yml --ignore tests",
-    "prestart": "npm run build:production",
-    "start": "cross-env NODE_ENV=production forever start --plain app.js",
+    "watch:cssjs": "cross-env NODE_ENV=development nodemon --ext css,js --ignore tests --ignore public/assets/css/critical.css --ignore public/assets/js/main.dev.js --ignore public/assets/js/bundle.dev.js --exec \"npm run build:dev\"",
+    "watch:other": "cross-env NODE_ENV=development nodemon app.js --ext pug,svg,png,yml --ignore tests",
+    "start": "npm run build:prod && cross-env NODE_ENV=production forever start --plain app.js",
```
jmervine: I wouldn’t add ‘NODE_ENV=production’ here. That should be picked up from the environment. 

I’d also consider having ‘npm build’ switch on NODE_ENV to run build:prod or build:dev, making start work for both development and production.

jmervine:
Guessing the tests are failing because critical is running before the app finishes starting up.

package.json
XhmikosR: This was already here (the `NODE_ENV` setup). And in fact we did this from the shell.js era too.

I still need to rethink the scripts part and there's definitely room for improvement. I'll ping you if I don't manage to find a good solution.

XhmikosR:
You mean the tests on AppVeyor? If so, I'm not sure how to work around it in a cross-platform way currently.

XhmikosR:
@jmervine: here's the situation:

1. `npm run dev`/`npm run watch` is broken
2. Heroku is broken and doesn't seem to respect the Procfile's `npm start`
3. The CI failures *I think* it's because of firewall policies. At least for AppVeyor, I think the firewall dialog opens up and fails to connect because `node.exe` isn't allowed to listen to that address/port. For Travis I assume it's something similar.

Locally, `npm start` works properly for me, the same does `npm test`.

XhmikosR:
All right, so I can switch to nodemon maybe for the critical script.

This leaves us 1 and 3.

XhmikosR:
Er, `http-server`.

jmervine:
For 3, you can try running a different point (e.g. 3339) or something else less likely to barf. That said, running on localhost like you are, shouldn't be tripped up by firewall rules.

For 1, I'm going to need to look at what they do. 

At this point, I'd probably just pregenerate and check it in.

XhmikosR:
But how are we going to make sure we always update critical.css?

I don't mind including the file, I just want us to have to remember as few things as possible.

XhmikosR:
@jmervine:

* CI is fixed
* dev seems to work too
* Only issue left is Heroku

package.json
```diff
@@ -22,7 +22,7 @@
     "watch:cssjs": "nodemon --ext css,js --ignore tests --ignore public/assets/css/critical.css --ignore public/assets/js/main.dev.js --ignore public/assets/js/bundle.dev.js --exec \"npm run build:dev\"",
     "watch:other": "nodemon app.js --ext pug,svg,png,yml --ignore tests",
     "start-node": "node app.js",
-    "start": "npm run build:prod && cross-env NODE_ENV=production forever start --plain app.js",
+    "start": "npm run build:prod && cross-env NODE_ENV=production && node app.js",
```
XhmikosR: 1. Won't this break `npm stop` and `npm restart` scripts?
2. You don't need `&&` BTW
3. This won't run `build:prod` which is needed for the files to be generated...

jmervine:
@XhmikosR It's failing here
```
critical http://127.0.0.1:3339/ --base public --assetPaths public --ignore "@import" --ignore ":root" > public/assets/css/critical.css
```

I don't think don't think we should do these steps at all. Even if it means that we can't "async all the stuff".

```
> bootstrapcdn@0.0.1 critical-gen /Users/jmervine/Development/bootstrap-cdn
> cross-env GENERATE_CRITICAL_CSS=true PORT=3339 npm-run-all --race --parallel start-node critical                                                                                                              
             
                                                                                                                                                                                            
> bootstrapcdn@0.0.1 critical /Users/jmervine/Development/bootstrap-cdn
> wait-on tcp:3339 && critical http://127.0.0.1:3339/ --base public --assetPaths public --ignore "@import" --ignore ":root" > public/assets/css/critical.css                                                    
```

This is adding a ton of complexity.

Also, can you remove whatever installed `fsevents`, which in turn installed Chromium. Peeking at `package-lock.json`, I think it's `@babel/cli`. Not 100% sure what "babel" is used for (cross-compiling ES6 for client side applications I think?), but I'd like to find a workaround to avoid the `fsevents`.

XhmikosR:
fsevents was there. It just applied for macOS and they changed the behavior in npm so that it is included in package-lock.json, even if they are not installed to avoid cross-platform modifications.

Anyway just do `npm ls package` to see where eash dep is coming from.

Babel can't go away if we want to use ES6. And we were already using it regardless of this PR.

I can commit the file, instead of doing this on the fly, but:

1. develop should be already broken too on Heroku because of the JS files
2. how can we make sure the preprocessed assets are always up to date before a release?

XhmikosR:
@jmervine: please check now, I committed the assets and everything works.

We just need to find a way to make sure we don't miss committing the assets when they change. Maybe a hook on Travis to auto-commit them?

jmervine:
I think GitHub supports precommit hooks. We could look in to that and add it to the dev docs.

XhmikosR:
Well, I could the same for you, now, couldn't I?
Everything should be cross-platform, because not everyone is using the OS you or I use.

Sorry, can't remove Chromium because **I** don't add it. It's how critical works, which is how puppeteer works.

XhmikosR:
I'm gonna close this mostly because I don't want to argue more. To me, this PR should be merged because the benefits are huge, but anyway.

PR #988, opened 2018-03-27
Replace "MaxCDN" with "StackPath".

XhmikosR:
Only thing left is the GitHub organization and the Travis notification email.

PR #990, opened 2018-04-04
heroku ci

jmervine:
This is based off of https://github.com/MaxCDN/bootstrapcdn/pull/988.

It removes AppVeyor and Travis. I updated Travis settings to not run if `.travis.yml` is not present, however, I don't have the access to do the same with AppVeyor (if it supports it).


jdorfman:
@jmervine do have to get rid of Travis? I like that it is public, were Heroku is hidden.

jmervine:
@jdorfman I _think_ heroku tests runs are public as well, I'd have to log out to check.

PR #996, opened 2018-04-06
Clean up .gitignore and config/_maxcdn.yml.

XhmikosR:
I'm not 100% sure if any of these are still used, so @jmervine please confirm.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,14 +1,9 @@
 *.log
 .DS_Store
-.popular.json
 .zedstate
 config/_config.yml.bak
-config/_maxcdn.yml
 Dockme.yml
 lint.html
 npm-debug.log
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -1,14 +1,9 @@
 *.log
 .DS_Store
-.popular.json
 .zedstate
 config/_config.yml.bak
-config/_maxcdn.yml
```
jmervine: Please don't remove this.

XhmikosR:
So this is still used, right? If so, I'll keep the config file I removed
too.

.gitignore
jmervine: Oh -- maybe not. I thought I did. Never-mind.

PR #1002, opened 2018-04-10
Remove unneeded callback.

XhmikosR:
Leftover from 4dad3f9.

@jmervine: I'm trying to split as many things as possible from #1000 so that it's easier to review.


jmervine:
👍🏽

PR #1004, opened 2018-04-10
Clean up content types.

XhmikosR:
TODO: revisit ttf and otf mime types.


jmervine:
Might consider removing the logic that supports multiple content types if we’re removing the additional ones.

XhmikosR:
@jmervine: yup, definitely worth it. Especially now we **do** specify them properly.

I might have a go at it in the future unless you beat me to it.

PR #1009, opened 2018-04-10
Use the same version's CSS/JS files.

XhmikosR:
This is just so that we are safe and we don't accidentally load incompatible versions.

So, now, the Bootstrap JS file loaded will be the same as the Bootswatch version.

### DIFF
--- a/app.js
+++ b/app.js
@@ -223,6 +223,7 @@ app.locals.helpers = helpers;
 app.locals.config = config;
 app.locals.basedir = PUBLIC_DIR;
 app.locals.getVersionedPath = staticify.getVersionedPath;
+app.locals.semver = semver;
 
 // routes
 app.get('/fontawesome/', routes.renderFontawesome);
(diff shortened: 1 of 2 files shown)


XhmikosR:
BTW, this mostly happens if Bootswatch isn't updated after a new release.

So, this way we will load the same Bootstrap version as the Bootswatch version we load.

The opposite shouldn't affect anything I believe.

XhmikosR:
Ideally I could try using `semver` in the templates. I'm just not sure if it's worth it, nor the performance impact etc.

But that would simplify things and would be the safest.

So, let me know and I could try the semver approach.

jmervine:
Approved the change. Didn’t realize the tests weren’t passing until after. Should reconcile that.

XhmikosR:
The tests failure might be unrelated. I'll have  a look before merging,
don't worry.

So you agree with this approach and using semver?

The reason for this is because there are cases when Bootswatch hasn't been
updated yet and there might be incompatibilities.

PR #1015, opened 2018-04-11
Nuke nodemon and switch to chokidar.

XhmikosR:
TODO:

- [ ] fix `dev` script; the port is already used on reload.

Reason for this change: too many deps and chokidar seems to do the job fine too as long as we fix the reload issue.


XhmikosR:
OK, I don't think I can make this work properly unfortunately.

PR #1018, opened 2018-04-12
Fix `Uncaught AssertionError`.

XhmikosR:
Note that this happens before my recent changes too.

I'm not sure if there are any other such cases, so if you spot one let me know.


XhmikosR:
BTW, I'm not an expert here, so I don't know if there's any other "proper" way to handle this. I used the good old `try...catch`.

jmervine:
If it works, it's proper enough. That said, I don't know of a better way off hand.

PR #1021, opened 2018-04-13
Remove `format` devDependency.

XhmikosR:


XhmikosR:
OK, this isn't ready; I need to fix the `assertContentType` calls too.

PR #1022, opened 2018-04-13
Tighten mocha options and move them to mocha.opts file.

XhmikosR:
Also, renamed the "tests" folder to "test" since that's where mocha is looking for.

@jmervine: there's a bug in our tests, with wrong callbacks in `assertValidHTML` thus why we need to use `--exit` in mocha. I could use some help fixing this. Also, the validator supports `ignore` so I say we drop this custom code.


XhmikosR:
I made #1027 which seems to be the culprit with no exit in mocha.

Thanks for the changes @jmervine

XhmikosR:
@jmervine: you can remove `--exit` from mocha.opts, make a test fail and see the hang.

But if you run `mocha:functional` which doesn't run the server, you will see the tests exit without a hang.

PR #1023, opened 2018-04-13
Clean up `CONTENT_TYPE_MAP`.

XhmikosR:
TODO:

- [ ] Change content types for ttf, woff and woff2 to `font/type`?
- [ ] Adapt tests so that we don't test content-type twice: one for `content-type: undefined` and a second for each file type manually

### DIFF
--- a/test/test_helper.js
+++ b/test/test_helper.js
@@ -14,7 +14,6 @@ const yaml       = require('js-yaml');
 
 let response = {};
 
-// For array of types, the first one will be chosen when testing strictly
 const CONTENT_TYPE_MAP = {
     css: 'text/css; charset=utf-8',
     js: 'application/javascript; charset=utf-8',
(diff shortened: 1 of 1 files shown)


XhmikosR:
@jmervine: I enabled mocha's full output.

As you can see, we have a lot of duplicate stuff in functional:

1. We need to remove `content-type: undefined` if we are going to check individually later
2. We don't need to run `assertItWorks both for each group test and in `public/**/*.*`
3. We probably should remove the custom code that stops testing `twitter-bootstrap` now that tests are fast
4. `content-type` in `public/**/*.*` is duplicate for most files since we do it individually too

I suggest that

1. we stop running functional explicitely for each group of files
2. change the `assertHeaders` some way so that the `content-type` is tested

Maybe I'm missing something, but I believe we could reduce the tests a lot. :)

XhmikosR:
@jmervine: I think you missed my comments here and approved by just looking at the code :)

jmervine:
You’re welcome to refactor as you see fit. I’ll review. Happy to take a stab but it won’t be in the next week to 10 days.

XhmikosR:
No problem, I'll try to figure it out, otherwise I'll wait. I definitely think this is worth it because we duplicate many tests currently.

XhmikosR:
I'm gonna merge this and make an issue for the other things.

PR #1024, opened 2018-04-13
Add Adobe Dreamweaver to Integrations

jdorfman:
cloudapp annotation 2018-04-13 at 7 25 08 am png


XhmikosR:
Sec, let me compress the image and revert the npm i patch

PR #1026, opened 2018-04-13
Cleaning up assertValidHTML

jmervine:
@XhmikosR Cleaned up the `assertValidHTML` method a bit. Created this PR on to your branch to you can just pull it in -- or close this and cherry-pick the commit if you prefer.

Example output with a forced error:
```
  1) valid html:
     Error: HTML Validation for '/' failed with:
        The “center” element is obsolete. Use CSS instead.
        From line 224, column 11; to line 224, column 18
        There were errors.
```

RE `--exit`. I remember adding that, but I wasn't able to reproduce the condition which caused mocha to hang. And insights on that would be helpful.

### DIFF
--- a/test/test_helper.js
+++ b/test/test_helper.js
@@ -96,12 +96,11 @@ function assertValidHTML(res, done) {
 
     validator(options, (err, data) => {
         if (err) {
-            console.trace(err);
             return done(err);
         }
 
-        // Returned when successful.
-        if (data.indexOf('The document validates') > -1) {
+        // Return when successful.
+        if (data.includes('The document validates')) {
             return done();
         }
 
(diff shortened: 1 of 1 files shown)


XhmikosR:
@jmervine: about the callback issue: edit layout.html and change `meta charset` line to read `utf-88` for instance. Run `npm t` and this will make tests hang.

XhmikosR:
I pushed a patch which should demonstrate the hang.

XhmikosR:
Hmm, actually the hang happens when something else errors out too.

I think we just keep using `--exit` for now and we can get back to it later. Maybe it's because we run the server but we don't close it?

XhmikosR:
That must be the reason, we don't close the server. Because `npm run mocha:functional` terminates just fine.

So we need to properly load the server and close it for testing. But that's for another issue. I'll make one to track it.

PR #1031, opened 2018-04-14
Add tests for sitemap.xml and robots.txt.

XhmikosR:


app.js
```diff
@@ -278,7 +278,7 @@ const map = sitemap({
     cache: 60000,       // enable 1m cache
     route: {            // custom route
         '/': {
-            disallow: NODE_ENV !== 'production' || false
+            disallow: !(NODE_ENV === 'production' || NODE_ENV === 'test') || false
```
jmervine: At this point perhaps just change it to !== development

app.js
```diff
@@ -298,7 +298,7 @@ const map = sitemap({
     }
 });
 
-if (NODE_ENV === 'production') {
+if (NODE_ENV === 'production' || NODE_ENV === 'test') {
```
jmervine: Perhaps just always render the site map? Is there a reason to have it not generated in dev?

XhmikosR:
I had to do this back then because you were using `development` for the Heroku deployments for PRs/dev subdomain.

If we don't do that anymore, then I can simplify the check for development only.

And yeah, you are right. The sitemap shouldn't hurt us; robots.txt should have disallow when not in production to prevent indexing. This includes the dev subdomain and PRs (although for the latter it shouldn't matter since they instances are destroyed IIRC).

PR #1033, opened 2018-04-16
Convert _config.yml to JSON.

XhmikosR:
This allows us to use it without any packages.


XhmikosR:
That doesn't make any sense to me.
Why not use what is natively supported?

I don't see how it's not readable, but I guess that is subjective.

How about pure JS then?

jmervine:
Pure JS would be better than JSON, but still not as good a YAML, which is the ideal configuration format regardless of language.

1. YAML is simple and readable.
1. YAML is less error prone and more forgiving.
1. YAML support comments.
1. YAML removes the necessity for extraneous syntax, e.g. semi-colons and open/close brackets.

PR #1034, opened 2018-04-16
 Make content-type checks stricter. 

XhmikosR:
@jdorfman @jmervine: please make sure you remove the files I removed here from the CDN too.

Previously we didn't check PNG files for the proper content-type. So IMHO, better be stricter than miss something.

PS. please also remove the files from #977, https://github.com/MaxCDN/bootstrapcdn/pull/977/commits/d85b5c971e8d67f8d73652bdcfd8dd68d04f995c so that we are in total sync. I don't have access to the stats, but the Bootstrap files I removed there were the wrong version. For the jQuery file, it just shouldn't have been there, but please verify the stats. If we are going to keep it on the CDN, we should commit it.


XhmikosR:
I removed them from GitHub but I can still access jQuery at least that I
tried on the CDN. So I don't think they have been removed from S3.

jdorfman:
@jmervine can we re-run the sync script to remove the `less` and `scss` files? Thanks

jmervine:
@jdorfman The script very intentionally doesn't remove anything from s3, only copies things to it. We've discussed and agreed upon in the past that we don't remove things from s3 as a general rule, as it hurts nothing to have it there. 

I guess the downside we didn't consider is that if we try to reconcile github w/ s3, using s3 as the source of truth, then we're pulling in old files.

I can make a version that is designed to remove files that don't exist in github, but unless we're to change our practice and agree that github is the source of truth (I'm in favor of this btw), when I don't want to flat sync from github to s3.

Down to discuss more if you'd like.

XhmikosR:
OK, as long as you guys don't add them back, we should be OK.

Tests will fail anyway, so you'll notice hopefully.

PR #1039, opened 2018-04-17
Remove async dep.

XhmikosR:
We only use it in tests.


XhmikosR:
@jmervine: I just pushed another patch which actually removes async. It seems we can use a simple `for...of` loop instead.

PR #1042, opened 2018-04-18
Make authors test stricter and move common code to test_helpers

XhmikosR:


XhmikosR:
I'm gonna merge this since I don't see something wrong. If you see something please ping me.

PR #1049, opened 2018-04-20
Fix endpoint adding a trailing slash for URIs with extension.

### LINKED ISSUE #1047: Fix `cleanEndpoint` for URLs with extension
```
GET /data/bootstrapcdn.json/ 200 3.130 ms - -
GET /robots.txt/ 200 1.406 ms - 25
GET /sitemap.xml/ 200 1.266 ms - 740
```

XhmikosR:
Fixes #1047


test/test_helpers.js
```diff
@@ -37,9 +37,9 @@ function getExtension(str) {
     // So, the result we want is the third Array element,
     // since the first one is the whole match, the second one
     // returns the first captured match, etc.
-    const re = /(\.)([a-zA-Z0-9]+)$/;
+    const match = str.match(/(\.)([a-zA-Z0-9]+)$/);
 
-    return str.match(re)[2];
+    return match && match[2];
```
jmervine: Will this return the value of match[2]? I’d assume it would return a boolean.

XhmikosR: You'd think so but it's not :)

The use of `&&` there takes care of it.

Example from when running the endpoints:

```
null
null
null
null
json
null
null
null
null
txt
null
xml
```

And for the functional tests, it just returns the extension as usual, otherwise tests would fail there.

XhmikosR: So basically it's like saying, if match isn't null then return match[2].

XhmikosR:
I'm going to merge this since it seems fine to me.

PR #1053, opened 2018-04-20
Add a custom 404 page.

XhmikosR:
Fixes #1028

PR #1056, opened 2018-04-20
tests: switch to `assert.ok` and `assert.strictEqual`.

XhmikosR:
`assert` is just an alias to `assert.ok` anyway.

**Should be merged after #1053 is merged.**


XhmikosR:
Ping @jmervine

PR #1060, opened 2018-04-21
main.js: switch back to `for` loops.

XhmikosR:
Unfortunately, IE 10 doesn't support this.


XhmikosR:
An alternative solution would be to either

1. include a polyfill
2. just assign `NodeList.prototype.forEach` to `Array.prototype.forEach`:
    ```js
    if (window.NodeList && !NodeList.prototype.forEach) {
        NodeList.prototype.forEach = Array.prototype.forEach;
    }
    ```

XhmikosR:
I went with plain for loops for now.

PR #1065, opened 2018-04-24
Pass the whole response to assertItWorks.

XhmikosR:


XhmikosR:
I'm not sure if this is better or not. In theory it might be worse so I'm gonna drop this patch.

PR #1067, opened 2018-04-24
Tighten `content-encoding` tests

### LINKED ISSUE #1066: Fix tests to only check for `content-encoding` for supported types

XhmikosR:
Fixes #1066


XhmikosR:
Test fail correctly because we have files (PNG files) on the CDN with gzip enabled.

After we get the SP guys to update the CDN, I'll restart tests. Good thing I was getting a few other failures that got me to look into this closer.

PR #1072, opened 2018-04-25
tests: cache staticify calls.

XhmikosR:
On a test machine, `npm run mocha` goes from ~1 min to 13 sec.


XhmikosR:
I'm going to merge this since it's OK and the speedup is real :)

PR #1088, opened 2018-05-01
Improve data tests.

### LINKED ISSUE #939: Improve data tests
So, that we catch it if it's not generated again.

XhmikosR:
@jmervine: this works apart from the `timestamp`.

Feel free to push any fix or propose any solutions.

Fixes #939.

### DIFF
--- a/lib/helpers.js
+++ b/lib/helpers.js
@@ -1,8 +1,9 @@
 'use strict';
 
-const fs   = require('fs');
+const fs = require('fs');
 const path = require('path');
-const sri  = require('sri-toolbox');
+const semver = require('semver');
+const sri = require('sri-toolbox');
 const yaml = require('js-yaml');
 
 const CONFIG_FILE_PATH = path.resolve(__dirname, '../config/_config.yml');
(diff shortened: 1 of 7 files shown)


lib/helpers.js
```diff
@@ -45,9 +46,36 @@ function getConfigPath() {
     return CONFIG_FILE_PATH;
 }
 
+function generateDataJson() {
+    const config = getConfig();
+    const data = {
+        timestamp: new Date(),
```
jmervine: [Here's a thread on mocking out `new Date()` in node](https://stackoverflow.com/questions/28504545/how-to-mock-a-constructor-like-new-date?utm_medium=organic&utm_source=google_rich_qa&utm_campaign=google_rich_qa). This is kind of what I was suggesting. Another option would be to use `Date.parse` here, perhaps, something like `Date.parse('04 Dec 1995 00:12:00 GMT');`.

XhmikosR:
@jmervine: I used `mockdate` and it seems to do the job.

Also note that I needed to add `toString()`.

XhmikosR:
OK I changed to `toISOString()` so it should be the same as before.

PR #1089, opened 2018-05-01
Add tests for redirects.

### LINKED ISSUE #1045: Test redirects too

XhmikosR:
@jmervine: The problem with this approach is that I'm not sure how to access the original request to test for a `301` response status code and for the target uri.

The only thing I've found so far is the `response`'s object which contains this bit we might be able to use:

```js
IncomingMessage: {
    request: {
        Request: {
             headers: {
                referer: 'http://localhost:3334/legacy/'
            }
        }
    }
}
```

Fixes #1045.

### DIFF
--- a/test/redirects_test.js
+++ b/test/redirects_test.js
@@ -1,0 +1,37 @@
+'use strict';
+
+const assert = require('assert');
+const helpers = require('./test_helpers.js');
+
+const config = helpers.getConfig();
+
+describe('redirects', () => {
+    const redirects = {
+        '/legacy/': '/legacy/bootstrap/',
+        '/alpha/': '/',
+        '/beta/': '/'
+    };
+
+    for (const redirectFrom in redirects) {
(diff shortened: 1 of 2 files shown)


jmervine:
Looking at `request`'s docs, what you suggested above is viable, otherwise you could either disable redirects, which might be good, as it would prevent a test from succeeding with an unintentional redirect.

You should be able to do that but updating [these options](https://github.com/MaxCDN/bootstrapcdn/blob/develop/test/test_helpers.js#L134-L138), adding
```
followRedirect: false
```

XhmikosR:
I already had tried that like 10 days ago but I think it didn't work.

https://github.com/MaxCDN/bootstrapcdn/compare/develop-xmr-request-options

I'll try with that enabled again.

jmervine:
Gotcha. If it's not working, then fall back on `referer` and perhaps I'll muck with disabling redirects later.

XhmikosR:
@jmervine: I believe I fixed it now. I'm also checking the target uri now apart from the `301` status code.

PR #1092, opened 2018-05-01
Add more info about authors.

XhmikosR:


views/about.pug
```diff
@@ -21,6 +21,13 @@ block content
                     li
                         a(href=author.url, rel='noopener', target='_blank')=author.name
 
+                        if author.work
+                            |  - 
+                            if author.work.text && author.work.url
+                                a(href=author.work.url, rel='noopener', target='_blank')=author.work.text
+                            else if author.work.text
+                                | #{author.work.text}
+
```
jmervine: Since `work.text` appears to be required in all cases, perhaps this would be simpler
```
if author.work && author.work.text
    | -
    if author.work.url
        a(href=author.work.url, rel='noopener', target='_blank')=author.work.text
    else
        | #{author.work.text}
```

XhmikosR: Yup, you are right. Initially I added the YAML code as an array hence this leftover.

I'll change it before merging.

PR #1095, opened 2018-05-02
Move redirects to config

### LINKED ISSUE #1094: Move redirects to a json config file

XhmikosR:
Fixes #1094 

Later, this will be used in routes too.

### DIFF
--- a/config/_config.yml
+++ b/config/_config.yml
@@ -31,6 +31,16 @@ javascript:
     uri: /assets/js/vendor/clipboard.min.js
   - name: main
     uri: /assets/js/main.js
+redirects:
+  - name: legacy
+    from: /legacy/
+    to: /legacy/bootstrap/
+  - name: alpha
+    from: /alpha/
+    to: /
+  - name: beta
+    from: /beta/
+    to: /
 bootstrap:
   - version: 4.1.1
     current: true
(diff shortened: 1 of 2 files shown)


jmervine:
This feels like an unnecessary layer of abstraction.

jmervine:
Also, if you're going to abstract these, we should add a section to https://github.com/MaxCDN/bootstrapcdn/blob/develop/config/_config.yml

XhmikosR:
I don't want to use _config.yml. It's already bloated with CDN stuff and www stuff shouldn't be there.

Like I said, the plan is to use this in routes too. And later add each page title and description to a config file too.

XhmikosR:
I'll see how it'll look like in YAML. I just don't like having www and CDN stuff in one file because it has become pretty big.

XhmikosR:
@jmervine: please check again.

jmervine:
I personally like that better. I don’t disagree that _config.yml is getting a bit overloaded, but unless we’re going to throughly break them apart and load them in to a centralized config object, this works. 

I have strong opinions, weakly held, here. If you feel strongly about it being the other way, I’m down to discuss. Additionally if/when we break it out, I’m willing to consider using js over yaml, but I’d prefer to stay away from json.

XhmikosR:
Nah, I don't mind currently but eventually we should break config.

So, this works for you?

PR #1098, opened 2018-05-05
Clean up tests more.

XhmikosR:
* Reduce the number of variables we are passing around in tests
* Move variables in the block they are used
* Use `assert`'s strict mode
* Add back a try/catch block in `assert.itWorks` to fix the uncaught assertion error
* Remove unneeded global variable `responses` in test_helpers.js
* Move `domainCheck` in functional since it's only used there
* Remove unused parameter and code from `assertHeaders`
* Rename `preFetch` to `prefetch`
* Remove mocha's retries again; tests should not fail

### DIFF
--- a/test/404_test.js
+++ b/test/404_test.js
@@ -1,16 +1,14 @@
 'use strict';
 
-const assert    = require('assert');
-const helpers   = require('./test_helpers.js');
-
-const config    = helpers.getConfig();
-const uri       = helpers.runApp(config, '404');
-
-let response    = {};
+const assert = require('assert').strict;
+const helpers = require('./test_helpers.js');
 
 describe('404', () => {
+    const uri = helpers.runApp('404');
+    let response = {};
+
     before((done) => {
-        helpers.preFetch(uri, (res) => {
+        helpers.prefetch(uri, (res) => {
             response = res;
             done();
         });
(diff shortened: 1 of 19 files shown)


XhmikosR:
@jmervine: there's an issue I could use some help.

`request` in functional_test.js is broken when an error is thrown in `helpers.preFetch`. That is why we keep getting those "Uncaught assertion errors" in `assert.itWorks` in functional.

Also, I don't seem to call the callback at all in functional's `request`.

XhmikosR:
@jmervine: the issue I mention [above](https://github.com/MaxCDN/bootstrapcdn/pull/1098#issuecomment-386829213) still exists though. That is what I need help with :P

XhmikosR:
So basically here's what I was thinking:

We should drop `request` in functional and move the responses caching in test helper prefetch. Then run functional walk first, then run the individual tests there.

Also, perhaps we should return an object with the responses in test helper, although I'm not sure this would help us. In theory it seems more appropriate but it's just an idea.

XhmikosR:
@jmervine: I just went back to using a try/catch around `itWorks`.

Please check the diff again one last time.

XhmikosR:
@jmervine: ping ^^

PR #1103, opened 2018-05-07
Adding cov report to end of tests

jmervine:
@XhmikosR This is (barely) a start, pretty weak without a check against existing percent, but I wanted to get the report added, at least.


package.json
```diff
@@ -15,15 +15,16 @@
...
+    "test": "npm run lint && npm run mocha && npm run cov",
     "purge-latest": "bash ./scripts/purge.sh",
     "bootlint": "node ./scripts/update_bootlint.js",
     "bootstrap": "node ./scripts/update_bootstrap.js",
     "bootswatch": "node ./scripts/update_bootswatch.js",
     "integrity": "node ./scripts/integrity.js",
     "sri": "node ./scripts/sri.js",
-    "wp-plugin": "node ./scripts/wp-plugin.js"
+    "wp-plugin": "node ./scripts/wp-plugin.js",
+    "cov": "istanbul cover app.js"
```
XhmikosR: I think we should cover everything apart from public dir.

.eslintignore
```diff
@@ -1,4 +1,5 @@
 /.git/
+/coverage
```
XhmikosR: nitpicking but I prefer to use a trailing slash for folders to quickly distinguish them from files.

XhmikosR:
Something seems wrong here too with package-lock.json and it should be reverted.

I don't get why this happens for you though.

XhmikosR:
I'm gonna merge this since it's a good step to the right direction; we can tweak it later as needed.

PR #1115, opened 2018-05-11
Load Bootswatch from npm too.

### LINKED ISSUE #948: Change bootswatch.js script to use GitHub tags

XhmikosR:
We might need to try another module since fs-extra doesn't seem to cover us here; exclude still copies all the files...

Any help is welcome.

Fixes #948.

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,13 +79,18 @@ Contains the CDN files we host.The SRI values are updated by running `npm run in
 
 Our CSP config using <https://github.com/helmetjs/csp>
 
-## Updating Bootstrap/Bootlint
+## Updating Bootstrap/Bootlint/Bootswatch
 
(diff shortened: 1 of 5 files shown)


jmervine:
Might just be worth walking and only copying files that match the extensions we support.

XhmikosR:
@jmervine: can you review again? Also, please try this branch too just to be safe.

jmervine:
Resolved conflicts as well.

XhmikosR:
@jmervine: we'll need to try your suggestion later. :)

PR #1117, opened 2018-05-18
Twitter: switch to summary card.

XhmikosR:


XhmikosR:
Before:

1-before

After:

2-after

XhmikosR:
The test failures is the issue I have reported on SP without any progress yet.

Issue #1109, opened 2018-05-08
Make deployments more foolproof


XhmikosR:
@jmervine: it's your area of expertise since you only have access there. There must be some way we make a retry or something to prevent deployments not happening.

Because, I just happened to notice it and this could be very problematic.

jmervine:
Let me know next time this happens and I'll check the build history and see if heroku even got notified. If it did and failed it should be posted on #notifications in Slack, if not, then it's a breakdown either in github sending the webhook, travis notifying that it succeeded or heroku actually getting the notifications. Regardless, as mentioned in slack, we can can't alert or retry when no action is recorded. 

I can always manually trigger a build from the heroku dashboard via my phone, worst case.

XhmikosR:
I thought the failure was in `npm i` step or something hence why I believe there's room for improvement. Maybe a retry after an interval just to work around failures in this step.

If the hook failed for some reason, there's not a lot we can do, I get that.

XhmikosR:
@jmervine: did you disable the CI check requirement for deployments? If so maybe we can close this.

PR #1120, opened 2018-05-19
Production release

XhmikosR:


XhmikosR:
Note that I'm not merging this because currently gzip is broken completely.

XhmikosR:
@jmervine: production is again broken, i.e. not the latest master.

Can you look into improving this? I have an issue open for this #1109

PR #1126, opened 2018-05-21
Fix sitemap generation.

### LINKED ISSUE #1125: Fix sitemap.xml
This was caused by the routes split. I remember I was having issues moving sitemap and robots to their own modules. We'll need to either revert the routes modularization or figure out how to make this work right.

XhmikosR:
I'm not sure if this is the optimal solution. I definitely not like duplicating the route paths. Any ideas how to simplify this and make it less error-prone are welcome!

On a side note, our tests didn't catch this. Suggestions are welcome for this too.

Fixes #1125.


XhmikosR:
Yup, that was my idea too, but I didn't have the time to look into it more. I suspect we can.

Feel free to take a stab at it.

PR #1129, opened 2018-05-21
Add Coveralls support.

### LINKED ISSUE #1116: Add coveralls to the repo

XhmikosR:
Coveralls is added via OAUTH to my account.

The second commit is a live proof of code coverage :)

BTW I have set the limit to 85% but maybe we should set this is nyc.

Also, I don't know, do you guys have access to the Coveralls account? If not my guess is you'd need to login via GitHub, but I still don't know if you will have access to the Coveralls project settings. If not, it's a bummer, and generally I'm open to suggestions; I just went with Coveralls because I have used it in the past.

Fixes #1116

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,9 @@ node_js:
 
 script: npm run ci-with-coverage
 
+after_success:
+  - npm run coveralls
+
 matrix:
   fast_finish: true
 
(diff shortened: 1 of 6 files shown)


XhmikosR:
The second commit is a live proof of code coverage :)

BTW I have set the limit to 85% but maybe we should set this is nyc.

Also, I don't know, do you guys have access to the Coveralls account? If not my guess is you'd need to login via GitHub, but I still don't know if you will have access to the Coveralls project settings. If not, it's a bummer, and generally I'm open to suggestions; I just went with Coveralls because I have used it in the past.

XhmikosR:
Oh, and the AppVeyor failures are valid, one of the POPs serves ungzip'ed files...

PR #1138, opened 2018-05-25
Production release

XhmikosR:


XhmikosR:
I don't quite get why coveralls doesn't work

PR #1139, opened 2018-05-25
Fix navbar's active class.

XhmikosR:
Use a new variable based on `req.originalUrl`.

This was visible in the Jobs page where the Page title != link title. Not sure if there's another "proper" way to do this.


XhmikosR:
@jmervine I'm gonna merge this because we'll need it for #1141 too, but please let me know if you have a better solution later.

PR #1145, opened 2018-05-26
`bodyClass`: use `pageUrl` instead of the title.

### LINKED ISSUE #1142: make `bodyClass()` use the page url
Related to #1139

XhmikosR:
This broke with the page title was more complex and not just one word. Well, not broke, but wasn't what I'd expect.

Fixes #1142


XhmikosR:
Actually, it did break things leading to Error 500.

PR #1146, opened 2018-05-26
Move CDN files to /cdn/.

### LINKED ISSUE #1140: Move CDN files to `/cdn/`
The current situation is a bit messy IMO. Also, it shows us down, not a lot but it's noticeable in tests.

@jmervine @jdorfman what do you guys think? Preview: https://github.com/MaxCDN/bootstrapcdn/tree/develop-xmr-mv-cdn-files

XhmikosR:
Fixes #1140


XhmikosR:
Hey this change shouldn't break anything. Test failures aren't related to
the files move IIRC.

But maybe one of you guys needs to look into the symlink because thats the
only thing that can break.

XhmikosR:
Yup, the symlink is broken. Since I'm on Windows I need one of you guys to fix this and append the patch to mine.

develop: 21923 passing (38s)
this brandh: 14851 passing (34s)

XhmikosR:
OK, I think I got it right now. I used my Linux Mint VM.

So, as long as the tests number is `21923` we are all good.

XhmikosR:
Hmm, now they are even more :P

BTW there are 2 more symlinks in the repo apart from the bootstrap one.

XhmikosR:
I see your point now @jmervine. I hadn't even thought of that, well, because it shouldn't happen in the first place...

XhmikosR:
If we have a significant amount of traffic, can't we redirect those? I mean, as far as I can remember, we never exposed these endpoints. So if people used it, it's their fault, but I agree we should handle this gracefully.

jmervine:
@jdorfman That's a tough call. About 2/3 of our traffic is to cdn assets, but it's only about 100 per minute, so not a ton. Additionally, a lot of them are from the same folks, best I can tell. 

I'd lean towards merging this and breaking for those few, but I'm 🤷‍♂️

XhmikosR:
This needs resolving the conflicts, but if I do it on Windows, the symlinks will break.

Can one of you guys do it for me?

XhmikosR:
OK, I did the rebase on my Linux Mint VM.

PR #1147, opened 2018-05-26
Tweak theme query functions and add a test.

XhmikosR:


test/bootswatch4_test.js
```diff
@@ -39,6 +39,16 @@ describe('bootswatch4', () => {
         helpers.assert.bodyClass('page-bootswatch', response, done);
     });
 
+    it('handles invalid theme query parameter', (done) => {
```
jmervine: You could toss this inside of an array that tests for all the possible invalid options. 
```
[ -1, 500, 'foobar' ].forEach((i) => {
    it('handles invalid theme query parameter (' + i + ')', (done) => {
        // test here
    });
});

**Comment by XhmikosR** [comment]
@jmervine: can you look one last time? I improved the newly added test..

File: test/bootswatch4_test.js [hunk 2]
```diff
@@ -39,6 +39,29 @@ describe('bootswatch4', () => {
...
 
+    const totalThemes = config.bootswatch4.themes.length;
+    const invalidQueries = [
+        parseInt(-1, 10),
+        parseInt(0, 10),
+        parseInt(1, 10),
+        parseInt(500, 10),
+        parseInt('5', 10),
+        parseInt(totalThemes - 1, 10),
+        parseInt('foobar', 10)
```
jmervine: I would expect 0, 1 and Len-1 to return valid items in the array. They don’t?

XhmikosR: Yeah, they do. This just tests that an error is not thrown. Maybe we should tweak the test more to make it clear.

jmervine: I'll leave that to you. Currently it reads that all those values are invalid though.

XhmikosR: True, I removed those for now.

PR #1148, opened 2018-05-26
Production release

XhmikosR:


XhmikosR:
Yup, I mentioned this also to Justin. We need more content.

@jmervine: how do the appendLocals.js changes look to you? At first I had an issue with us overriding the whole `res` object the way I implemented so I ended up using `Object.assign`. It *should* be OK, but please confirm :)

jdorfman:
@jmervine I agree 100%, I will be working on the content now before pushing the Books and Jobs pages

PR #1149, opened 2018-05-27
Assorted tweaks

XhmikosR:


XhmikosR:
Nope, not possible without spending an unjustified amount of time. I personally don't find any value on this, but it was mostly an experiment anyway.

jdorfman:
LGTM

PR #1155, opened 2018-05-28
Improve tests

XhmikosR:
I didn't plan to merge this yet, but we need the extra debug info to investigate the `gzip` issue where sometimes it shows up and another it doesn't.

@jmervine: if you have any better implementation ideas please let me know.


XhmikosR:
@jmervine: before anything else, we need this so that we can provide enough debug info to the SP guys to solve the gzip failures.

Please review carefully

.travis.yml
```diff
@@ -6,6 +6,9 @@ language: node_js
 node_js:
   - "10"
 
+before_script:
+  - curl https://v4.ifconfig.co/
```
jmervine: What’s the purpose of this?

XhmikosR: So that we get the machine's IP for the SP guys.

XhmikosR:
1. provide X-HW when a test fails
2. provide the machine's IP
3. make sure the production headers don't have the debug headers and the opposite

jmervine:
Gotcha. Looks like a good approach.

PR #1156, opened 2018-05-29
Add more info about sustainability.

XhmikosR:


XhmikosR:
@jdorfman: I added one more patch which changes Sticker Mule URL.

PR #1161, opened 2018-06-05
Added gh issue template

jdorfman:


XhmikosR:
I'm not very familiar with the templates, but we'll see it in action :)

PR #1163, opened 2018-06-11
Update issue_template.md

XhmikosR:


jmervine:
What’s the purpose of this change?

XhmikosR:
The template is supposed to be a helper, not actually output to the issue :P

PR #1177, opened 2018-06-30
Updated About

jdorfman:


package.json
```diff
@@ -38,9 +38,11 @@
     "express-sitemap": "^1.8.0",
     "express-sslify": "^1.2.0",
     "helmet": "^3.12.1",
+    "jquery": "^3.3.1",
```
XhmikosR: Definitely revert the npm changes.

PR #1179, opened 2018-07-01
tests: decrease mocha's timeout

### LINKED ISSUE #1173: Reduce mocha's timeout
The only tests that might take a long time are the HTML validation ones. The CDN responses shouldn't take 15 seconds, which is what we currently set in mocha.

XhmikosR:
When we expect tests to take longer, set the timeout to 5 seconds explicitly. Otherwise use a 3 second timeout.

@jmervine: this might need to be tweaked a little more, but from my tests, it seems 5 seconds is enough.

Fixes #1173.

### DIFF
--- a/test/404_test.js
+++ b/test/404_test.js
@@ -18,7 +18,8 @@ describe('404', () => {
         done();
     });
 
-    it('valid html', (done) => {
+    it('valid html', function(done) {
+        this.timeout(helpers.TESTS_TIMEOUT);
         helpers.assert.validHTML(response, done);
     });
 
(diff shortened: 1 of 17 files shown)


test/404_test.js
```diff
@@ -18,7 +18,8 @@ describe('404', () => {
         done();
     });
 
-    it('valid html', (done) => {
+    it('valid html', function(done) {
+        this.timeout(5000);
```
jmervine: I'd pull the `5000` out in to a variable set in `test_helpers.js`, but I won't block on it.

XhmikosR: All right, I'll try to do it tomorrow.

PR #1181, opened 2018-07-01
Fix canonical URL and og:url.

### LINKED ISSUE #1180: Fix canonical URL and og:url

XhmikosR:
Fixes #1180

### DIFF
--- a/routes/appendLocals.js
+++ b/routes/appendLocals.js
@@ -63,7 +63,7 @@ function generateSRI(file) {
 
 function appendLocals(req, res) {
     const siteUrl = getCurrentSiteurl(req);
-    const canonicalUrl = `${config.siteurl}${req.path}`;
+    const canonicalUrl = `${config.siteurl}${req.originalUrl}`;
     const theme = getThemeQuery(req);
     const pageUrl = req.originalUrl;
     const bodyClass = generateBodyClass(pageUrl);
--- a/routes/appendLocals.js
+++ b/routes/appendLocals.js
@@ -1,4 +1,5 @@
 const path = require('path');
+const url = require('url');
 const helpers = require('../lib/helpers.js');
 const config = require('../config');
 
@@ -38,8 +39,8 @@ function getThemeQuery(req) {
     return query;
 }
 
-function generateBodyClass(url) {
-    let str = url;
+function generateBodyClass(pageUrl) {
+    let str = url.parse(pageUrl).pathname;
 
     if (str === '/') {
         str = 'home'; // only for the index page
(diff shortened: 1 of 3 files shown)


routes/appendLocals.js
```diff
@@ -63,7 +63,7 @@ function generateSRI(file) {
 
 function appendLocals(req, res) {
     const siteUrl = getCurrentSiteurl(req);
-    const canonicalUrl = `${config.siteurl}${req.path}`;
+    const canonicalUrl = `${config.siteurl}${req.originalUrl}`;
```
XhmikosR: I'm pretty sure this worked up some point.

XhmikosR:
I'll try to add tests for these tomorrow.

XhmikosR:
One thing left: should we distinguish pages based on the `theme` param? Because that's what we are doing.

routes/appendLocals.js
XhmikosR: OK, this broke after b0238932d9c0b9fa5c904cf990bdf29e7a814922

Inside say about.js:

```js
console.log(req.originalUrl);   // /about/
console.log(req.path);          // /
console.log(req.pathname);      // undefined
console.log(req.url);           // /
```

I'm not sure how to fix this and keep this modular structure...

views/_partials/head/social.pug
```diff
@@ -1,7 +1,7 @@
 meta(property='og:title', content!=title)
 meta(property='og:type', content='website')
 meta(property='og:description', content!=pageDescription)
-meta(property='og:url', content=siteUrl)
+meta(property='og:url', content=siteUrl + pageUrl)
```
jmervine: Shouldn't this be 
```
content=`${siteUrl}${pageUrl}`
```

routes/appendLocals.js
jmervine: Is it working as expected with this change to `req.originalUrl`?

XhmikosR: Yes, but `originalUrl` contains the query params which isn't what we want. I verified that this happens after  the aforementioned commit. For example, [this](https://github.com/MaxCDN/bootstrapcdn/tree/develop-xmr-before-modularize-app) branch doesn't have the canonical URL issue.

So, something is overwritten somewhere or something like that.

XhmikosR: If we don't find a solution soon, I'll need to revert the whole modularization and I don't want to do it :/

But I don't know how to solve this issue. I could possibly work around it by using `url.parse`, but I feel we are doing something wrong in the Router.

XhmikosR:
@jmervine: I added a workaround. I know I should add tests too but I don't have a lot of time. It should fix the issues though.

PR #1195, opened 2018-07-13
update the SP links to the new correct locations and add ref tracking

elof:
Hey all. Marketing is asking me to update various links we have out in the wild with some basic tracking. Please let me know if you have any questions at all.


XhmikosR:
Please revert any whitespace changes because those are need by the
templating language. Otherwise I'll take care of it tomorrow.

PR #1212, opened 2018-09-03
Switch to express-minify.

XhmikosR:
Note that I can't find a way for the SRI hashes to work for the minified files.

WIP until I test the revert of the clipboard SRI removal since this isn't being minified on the fly.


XhmikosR:
I'm gonna close this for now. I don't think it's worth it and I might even go back to inlining the whole main.js file.

PR #1222, opened 2018-10-08
Switch to CodeFund

### LINKED ISSUE #1221: Add codefund "bottom banner" ad code
Embed the following script into your website and create a `div` with `id="codefund_ad"`

```
<script src="https://codefund.io/scripts/c6a10903-6a7f-435b-92b2-99c9c9ce3379/embed.js"></script>
<div id="codefund_ad"></div>
```

It will look like this: 
image

via: https://sweetalert2.github.io/

This will add another ~$200 a month to our OC.

XhmikosR:
**DO NOT MERGE YET**

Fixes #1221 

#### CodeFund TODO:

* [x] the ads code needs to be async
* [x] we need a clear list of origins (scripts, images, etc) we then can whitelist in our CSP. So far I've only whitelisted a few things
* [ ] the tracking pixel (.gif) has the wrong MIME type
* [ ] the JS code has both ES5 and ES6+ code; it should be one thing consistently, IMO ES5 so that it works everywhere
* [ ] minify the JS code with careful settings for the browsers you want to support
* [ ] Enable gzip/brotli
* [ ] the PNG files have also the wrong MIME type
* [ ] no protocol-relative pixel should be loaded
* [ ] add https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Content-Type-Options where applicable
* [ ] add https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-XSS-Protection where applicable
* [ ] http/2
* [ ] minify the inlined CSS and templates code - low priority but no reason to send useless bytes
* [ ] lint the JS; there are global variables etc
* [ ] specify `max-age: 3600` or so for the JS file; it won't change faster than 3600s (1h) usually. The pixel and the JSON file should have `max-age: 0, must-revalidate` (and maybe `private`)


jdorfman:
We are going to replace this ad as well. Eric will be of assistance 

image 2018-10-08 at 10 30 16 am

coderberry:
@XhmikosR I will update the PR with the fix.

XhmikosR:
What fix? The async issue is know and I will work around it latwr but I'd
rather have the ads code work out of the box with async...

XhmikosR:
@coderberry: here's what I see so far:

1. the ads code needs to be async
2. we need a clear list of origins (scripts, images, etc) we then can whitelist in our CSP. So far I've only whitelisted a few things
3. the pixel image seems to load on http when the page is on http. Personally I'd only use https everywhere.

After the above, we can look into nuking CarbonAds.

XhmikosR:
A few more:

4. the tracking pixel (.gif) has the wrong MIME type
5. the JS code has both ES5 and ES6+ code; it should be one thing consistently, IMO ES5 so that it works everywhere
6. Again, no protocol-relative pixel should be loaded

coderberry:
@XhmikosR I made the changes to this PR: https://github.com/MaxCDN/bootstrapcdn/pull/1223

coderberry:
@XhmikosR 

- [x] the ads code needs to be async
- [x] we need a clear list of origins (scripts, images, etc) we then can whitelist in our CSP. So far I've only whitelisted a few things
- [ ] the pixel image seems to load on http when the page is on http. Personally I'd only use https everywhere. **This is built to support other publishers**
- [ ] the tracking pixel (.gif) has the wrong MIME type **Agreed. This was to solve a conflict we had. This will be replaced with an actual 1x1 gif, but not immediately. Is it a show-stopper?**
- [ ] the JS code has both ES5 and ES6+ code; it should be one thing consistently, IMO ES5 so that it works everywhere **Agreed. We will fix this as well**
- [ ] Again, no protocol-relative pixel should be loaded

XhmikosR:
1. doesn't work for me for the first visit when the JS isn't cached.
2. I commented on the PR but IMO this should be a subdomain, not a random generated CloudFront subdomain.

coderberry:
@XhmikosR Thank you for your feedback. This is awesome.

@hopsoft ^

XhmikosR:
7. Minify the JS code with careful settings for the browsers you want to support

jdorfman:
Great work @XhmikosR.

@coderberry look forward to seeing the changes =)

XhmikosR:
You should also enable gzip compression and http/2 @coderberry .

XhmikosR:
Maybe also add https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Content-Type-Options and https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-XSS-Protection

XhmikosR:
I added a few notes [above](https://github.com/MaxCDN/bootstrapcdn/pull/1222#issuecomment-427930715)

I managed to save another 15% from the uncompressed JS when minified, but it's untested ofc. IMO, doing 12, 11 and 6 will ensure the JS file is correct and as small as possible. So that along with gzip/brotli and http/2 will make sure the ads have the minimal impact on page load times.

XhmikosR:
@coderberry: I moved my suggestions to the first post of the PR. Blocking for me are the first 6 items, but you should really fix most if not all of these.

jdorfman:
@coderberry the `.png`'s also have incorrect mime types

```
> icurl https://cdn.codefund.io/21400d24-a68b-4619-9046-51190e723bde_consensys-academy-3-260x200.png
HTTP/2 200
date: Tue, 09 Oct 2018 16:33:25 GMT
content-type: application/octet-stream
```

Must be an misconfiguration on the origin that Cloudflare is just passing on.

XhmikosR:
The lint JS part shouldn't be taken lightly BTW. There are global variables leaks.

```
window.url_with_height_width
"https://codefund.io/t/s/c6a10903-6a7f-435b-92b2-99c9c9ce3379/details.json?width=2176&height=767"
```

XhmikosR:
We are going to merge this sooner due to some other reasons.

@coderberry: please keep the [TODO](https://github.com/MaxCDN/bootstrapcdn/pull/1222#issue-221189668) somewhere on your short-term TODO list :)

PR #1223, opened 2018-10-08
Replace Carbon Ads with CodeFund

coderberry:


views/_partials/codefund.pug
```diff
@@ -0,0 +1,3 @@
+.mt-5.mt-md-0.mb-lg-4
+    #codefund
+    script(async, src='https://codefund.io/scripts/c6a10903-6a7f-435b-92b2-99c9c9ce3379/embed.js?target=codefund&theme=bootstrapcdn')
```
XhmikosR: That won't work for the first visit.

config/helmet-csp.js
```diff
@@ -51,7 +51,8 @@ const CSP = {
         'codefund.io',
         'launchbit.com',
         'www.launchbit.com',
-        'www.ziprecruiter.com'
+        'www.ziprecruiter.com',
+        'd2nhukomolqgak.cloudfront.net',
```
XhmikosR: While this work, you should consider using a subdomain at some point.

views/layout.pug
```diff
@@ -50,7 +50,7 @@ html(lang='en', prefix='og: http://ogp.me/ns#', itemscope, itemtype='http://sche
                 .col-12.col-md-3
                     .row
                         .col-12
-                            include _partials/carbonads.pug
+                            include _partials/codefund.pug
```
XhmikosR: The old file needs to be removed.

config/helmet-csp.js
```diff
@@ -15,10 +15,7 @@ const CSP = {
         'platform.twitter.com',
         'cdn.syndication.twimg.com',
         'api.github.com',
-        'cdn.carbonads.com',
-        'srv.carbonads.net',
         'adn.fusionads.net',
```
XhmikosR: I think we can remove this one too.

jdorfman: what if we need to revert?

XhmikosR: We'll revert the whole patch anyway, so I'd rather have this complete.

XhmikosR:
I'll verify tomorrow with a clear head the async situation with a clear head.

PR #1230, opened 2018-10-30
Use destructuring for Objects and update deps

XhmikosR:


scripts/update_bootlint.js
```diff
@@ -4,7 +4,7 @@ const fs = require('fs');
 const path = require('path');
 const uglifyjs = require('uglify-js');
 
-let version = process.argv[2];
+let [, , version] = process.argv;
```
jmervine: This form seems far less clear or elegant than the previous. My 2 cents. 🤷‍♂️

scripts/sri.js
```diff
@@ -2,7 +2,7 @@ const path    = require('path');
 
 const helpers = require(path.resolve(__dirname, '../lib/helpers.js'));
 
-const digest  = helpers.sri.digest;
+const { digest }  = helpers.sri;
```
jmervine: This form seems far less clear or elegant than the previous. My 2 cents. 🤷‍♂️

XhmikosR:
I don't love Array destructuring either, but it's handy. That being said, I have no problem dropping that patch. Object destructuring does seem clearer to my eyes.

Also there is probably a way to just require destructuring for objects. I'll look into that if that's OK with you.

PR #1224, opened 2018-10-09
Nuke Travis and use AppVeyor for both Linux and Windows tests.

XhmikosR:
I started this mostly as an experiment, but it seems it is faster, at least per-job. In total, it's about ~5-6 minutes but this includes both jobs. AppVeyor doesn't offer parallel builds in the free plan as far as I can tell.

Still, I find this more stable so far from my tests.

So, if we decide to go with this, here's the TODO:

- [x] Fix conditional caching
- [ ] Switch to the GitHub app for the AppVeyor repo integration
- [ ] Remove Travis from the repo hooks/required PR checks
- [ ] Add Joshua in the AppVeyor account

/CC @jmervine @jdorfman

PR #1235, opened 2018-11-24
Nuke AppVeyor and use Travis for both Linux and Windows tests.

XhmikosR:
This is the opposite of #1224.

Opening this to see it in action and make a final decision.


XhmikosR:
Oh, also, in this patch I just run the full tests only on Windows since I can't find a way to enable symlinks on Windows for Travis.

Issue #1012, opened 2018-04-11
Add `Timing-Allow-Origin` header


XhmikosR:
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Timing-Allow-Origin

The header is already being used by many popular CDNs, like Google.

petarpetrovic:
I'll work on getting the header added sometime next week.

bripkens:
Thank you @XhmikosR!

PR #1241, opened 2018-12-04
Expose resource timings

bripkens:
By adding this header, users of the CDN can inspect a lot more details
their resource loading, e.g. track asset sizes.

See https://www.w3.org/TR/resource-timing-2/#sec-timing-allow-origin
for details.

Fixes https://github.com/MaxCDN/bootstrapcdn/issues/1012

### DIFF
--- a/app.js
+++ b/app.js
@@ -104,6 +104,8 @@ app.use((req, res, next) => {
     res.setHeader('Last-Modified', new Date().toUTCString());
     res.setHeader('X-Hello-Human', 'Say hello back! @getBootstrapCDN on Twitter');
     res.setHeader('X-Powered-By', 'StackPath');
+    // https://www.w3.org/TR/resource-timing-2/
+    res.setHeader('Timing-Allow-Origin', '*');
 
     res.locals.nonce = nonce;
 


XhmikosR:
This is useless because it doesn't affect the CDN.

XhmikosR:
And now I realize my reply isn't the best one; sorry I was doing multiple things.

bripkens:
What would need to be done in order to set this header? I identified this code due to the `X-Hello-Human` header which is available at that code location and when receiving assets via the CDN.

bripkens:
And now I realize that the cache-control headers are of course different :)

XhmikosR:
The CDN is a completely separate environment.

I could add the header there, I too think it's useful along with the Cache-Control header improvements.

I'll discuss this with the StackPath support before adding it.

bripkens:
Thank you!

bripkens:
Maybe they also wanna add that to the jQuery CDN btw. I noticed that they are operating both ;)

XhmikosR:
You are welcome to contact them about the jQuery CDN, I have done this many times with mixed results :/

The problem is that I couldn't find anyone from the jQuery team to confirm the changes, and I'm not the one that I can approve such changes.

bripkens:
Alright, thank you. I don't know the organizational structure behind the operation of these CDNs. So I hoped that there might only be a small team responsible on StackPath's side for the operation of open source CDNs :)

XhmikosR:
I remember for example asking them to enable http/2 for the jQuery CDN and it was blocked because I wasn't one of the jQuery people to approve the change.

PR #1261, opened 2019-01-08
Add abp test script into head

coderberry:


XhmikosR:
What exactly is this for? Also when a variable is global I'd rather be
explicit and use window. But I'm seriously interested about the extra stuff
you add which are synchronous so this is a no go for me.

coderberry:
Please disregard. We no longer need this in place.

This was for our ability to track uplift with eye/o (Ad Blocker whitelisting)

PR #1265, opened 2019-01-14
Lint

XhmikosR:
I kept the `url.parse()` changes as a separate patch but I'll squash before the merge.


XhmikosR:
Needs some more cleanup.

XhmikosR:
@jmervine please review one more time. I basically split `req.originalUrl` in `?` to keep only the pathname.

PR #1280, opened 2019-02-17
Production release

XhmikosR:


XhmikosR:
That's outside of this repo scope IMO.

You should always do `jpegoptim foo.jpg` before committing anything, everywhere.

PR #1282, opened 2019-02-25
Add Bootlint, Bootswatch 3 and Bootswatch 4 in bootstrapcdn.json.

XhmikosR:
Fixes #1281, fixes #1283

PR #1284, opened 2019-02-27
Add fully functional and dynamic API

### LINKED ISSUE #1281: JSON API missing bootswatch & bootlint
Hello again 😄

I originally opened #460 a while back and from that, y'all ended up creating:

https://www.bootstrapcdn.com/data/bootstrapcdn.json

This has been fantastic but given the recent changes/overhaul I've been doing in https://www.drupal.org/project/bootstrap/issues/3031415, I'm realizing this data is rather incomplete.

It doesn't contain any of the other package data that this CDN provides (e.g. bootswatch, bootlint).

I see that this JSON file is dynamically generated here:
https://github.com/MaxCDN/bootstrapcdn/blob/develop/lib/helpers.js#L37-L60

### LINKED ISSUE #1283: Revisit JSON entries
1. We don't list Bootstrap >= 4. Maybe we should revisit this? Not sure how breaking it can be
2. Add SRI hashes. This unfortunately will change the output format for Font Awesome, Bootlint and Bootswatch because we need a key for each one. So if we are going to go this way, better do it in #1282 

/CC @jdorfman @jmervine for thoughts

markcarver:
Fixes #1281, Closes #1282, Closes #1283

This will eventually replace the need for `config/_files.yml` and `scripts/integrity.js`.

The entire codebase should likely be refactored to utilize this new API data, but that is really follow-up work and a separate PR.

**New API documentation page:**
https://bootstrapcdn-dev-pr-1284.herokuapp.com/api/

**Deprecated existing "API" file**
https://bootstrapcdn-dev-pr-1284.herokuapp.com/data/bootstrapcdn.json

**/api/v1/:name**
https://bootstrapcdn-dev-pr-1284.herokuapp.com/api/v1/bootstrap
https://bootstrapcdn-dev-pr-1284.herokuapp.com/api/v1/bootswatch

**/api/v1/:name/:version**
https://bootstrapcdn-dev-pr-1284.herokuapp.com/api/v1/bootstrap/4.3.1
https://bootstrapcdn-dev-pr-1284.herokuapp.com/api/v1/bootswatch/4.3.1

### DIFF
--- a/lib/helpers.js
+++ b/lib/helpers.js
@@ -41,6 +41,8 @@ function getTheme(selected) {
 
 function generateDataJson() {
     const data = {
+        // @todo Figure out if/when this will go away?
+        deprecated: 'This file will no longer receive any updates. Visit https://www.bootstrapcdn.com/api for documentation.',
         timestamp: new Date().toISOString(),
         bootstrap: {},
         fontawesome: {}
--- a/lib/helpers.js
+++ b/lib/helpers.js
@@ -4,7 +4,7 @@ const fs = require('fs');
 const semver = require('semver');
 const sri = require('sri-toolbox');
 
-const { app, files } = require('../config');
+const config = require('../config');
 
 function generateSri(file, fromString) {
     file = fromString ? file : fs.readFileSync(file);
@@ -16,24 +16,24 @@ function generateSri(file, fromString) {
 
 function selectedTheme(selected) {
     if (typeof selected === 'undefined' || selected === '') {
-        return app.theme;
+        return config.app.theme;
     }
 
     const theme = parseInt(selected, 10);
 
     return theme === 0 || theme ?
         theme :
-        app.theme;
+        config.app.theme;
 }
 
 function getTheme(selected) {
-    const { themes } = files.bootswatch4;
+    const { themes } = config.files.bootswatch4;
 
     selected = selectedTheme(selected);
 
     return {
-        uri: files.bootswatch4.bootstrap
-                .replace('SWATCH_VERSION', files.bootswatch4.version)
+        uri: config.files.bootswatch4.bootstrap
+                .replace('SWATCH_VERSION', config.files.bootswatch4.version)
                 .replace('SWATCH_NAME', themes[selected].name),
         sri: themes[selected].sri
     };
@@ -42,13 +42,13 @@ function getTheme(selected) {
 function generateDataJson() {
     const data = {
         // @todo Figure out if/when this will go away?
-        deprecated: 'This file will no longer receive any updates. Visit https://www.bootstrapcdn.com/api for documentation.',
+        deprecated: `This file is deprecated and will no longer receive any updates. Visit ${config.app.siteurl}/api for documentation.`,
         timestamp: new Date().toISOString(),
         bootstrap: {},
         fontawesome: {}
     };
 
-    files.bootstrap.forEach((bootstrap) => {
+    config.files.bootstrap.forEach((bootstrap) => {
         const bootstrapVersion = bootstrap.version;
 
         if (semver.satisfies(semver.coerce(bootstrapVersion), '<4')) {
@@ -59,7 +59,7 @@ function generateDataJson() {
         }
     });
 
-    files.fontawesome.forEach((fontawesome) => {
+    config.files.fontawesome.forEach((fontawesome) => {
         data.fontawesome[fontawesome.version] = fontawesome.stylesheet;
     });
 
@@ -70,12 +70,10 @@ function generateDataJson() {
  * Helper methods used to handle the API.
  */
 const api = {
-
     badRequest(message = 'Bad Request.', suffix) {
         return this.message(message, suffix, 400);
     },
-
-    data(name, version = null, endpoint = config.apiVersion) {
+    data(name, version = null, endpoint = config.app.apiVersion) {
         const data = this.raw(name, endpoint);
 
         // Unknown data (without version).
@@ -105,12 +103,10 @@ const api = {
         // Valid version, return version specific assets.
         return data[version];
     },
-
-    latest(name, endpoint = config.apiVersion) {
+    latest(name, endpoint = config.app.apiVersion) {
         return this.tag('latest', name, endpoint);
     },
-
-    message(message, suffix = `Visit ${config.siteurl}/api for documentation.`, status = 200) {
+    message(message, suffix = `Visit ${config.app.siteurl}/api for documentation.`, status = 200) {
         if (suffix) {
             message += ` ${suffix}`;
         }
@@ -119,29 +115,24 @@ const api = {
             message
         };
     },
-
     notFound(message = 'Not found.', suffix) {
         return this.message(message, suffix, 404);
     },
-
-    raw(name, endpoint = config.apiVersion) {
+    raw(name, endpoint = config.app.apiVersion) {
         return (config.api[endpoint] && config.api[endpoint][name]) || null;
     },
-
     send(res, data, statusCode = null) {
         if (!statusCode) {
             statusCode = data.status ? data.status : 200;
         }
-        return res.status(statusCode).json(data);
+        res.status(statusCode).json(data);
     },
-
-    tag(tag, name, endpoint = config.apiVersion) {
+    tag(tag, name, endpoint = config.app.apiVersion) {
         const tags = this.tags(name, endpoint);
 
         return tags[tag] || null;
     },
-
-    tags(name, endpoint = config.apiVersion) {
+    tags(name, endpoint = config.app.apiVersion) {
         const tags = {};
 
         if (config.api[endpoint] && config.api[endpoint][name]) {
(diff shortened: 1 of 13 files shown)


markcarver:
@XhmikosR here ya go

scripts/api.js
```diff
@@ -0,0 +1,128 @@
...
+function getDirectoryNames(source, ignore = []) {
+    return fs.readdirSync(source).map((name) => ignore.indexOf(name) === -1 && name).filter((name) => {
+        if (!name) {
+            return false;
+        }
+        const absolute = path.join(source, name);
+        const stat = fs.lstatSync(absolute);
+
+        // @todo Remove symlink check when twitter-bootstrap files are moved
+        // to the proper directory.
```
XhmikosR: This cannot happen because of limitation of AWS.

scripts/api.js
```diff
@@ -0,0 +1,128 @@
...
+            const files = listFiles(versionPath);
+
+            files.forEach((file) => {
+                // Skip files that aren't considered assets.
+                if (!assetRegExp.test(file)) {
+                    return;
+                }
+                const stat = fs.lstatSync(file);
+                const relative = path.relative(versionPath, file);
+                const [, algorithm, hash] = sri.digest(file).split(/(sha\d{3})-(.*)/) || [];
```
XhmikosR: Currently we don't use Array destructuring anywhere on purpose. Not a big deal, just FYI.

views/api.pug
```diff
@@ -0,0 +1,133 @@
...
+                    sup &nbsp;
+                        a(href='#notes') 1
+
+            p
+                strong Example
+
+            p GET &nbsp;
+                a(href='/api/v1/bootstrap' target='_blank') #{config.siteurl}/api/v1/bootstrap
+
+            pre.hljs
```
XhmikosR: We don't use highlight.js.

XhmikosR:
@markcarver: thanks for the patch!

You should nuke the old scripts and files, if this is going to be merged too. And adapting all tests/scripts/etc too. But don't jump on this yet.

This requires to run the api script, similarly to our other scripts, and that's good.

My main concern is if we want to go this direction. Personally, what I don't like is that this is a change against going static and dropping express.js completely, which was my plan.

Secondly, rate limiting etc is something @jmervine needs to confirm on Heroku.

So, I'm in the middle personally, but I also don't have enough time this period to work on this apart from from simple straight-forward changes.

Let's wait for the others to chime in.

scripts/api.js
markcarver: Hm. Curious. What exactly is the limitation? I can update the code/commenting accordingly.

scripts/api.js
markcarver: This was to avoid creating another (unnecessary) `matches` variable just to extract the indexed values into other variables later. FWIW, the eslint rules prefer object destructuring, I felt this just followed suit.

jdorfman:
@XhmikosR @jmervine what if we set up a new pull zone dedicated to the API?

views/api.pug
markcarver: Sure. This was primarily just a byproduct of auto-converting from markdown to pug based on https://github.com/jsdelivr/data.jsdelivr.com/blob/master/README.md. I can remove the unnecessary class.

markcarver:
> You should nuke the old scripts and files, if this is going to be merged too. And adapting all tests/scripts/etc too. But don't jump on this yet.

Sure. I just wanted to keep this PR very specific and focus on the addition of a feature rather than clutter it with harder to review change lines. I felt we could address any migration/updating of code and removal of legacy scripts in a future PR.

> Personally, what I don't like is that this is a change against going static and dropping express.js completely, which was my plan.

Sorry, I'm not a mind reader. You asked for a PR and I gave one given this repo's current architecture. Regardless, it really wouldn't be that difficult to change in the future should that be the direction y'all want to go. The `api` script could simply spit out individual static JSON files rather than dumping everything into a YAML config file; just the same.

> Secondly, rate limiting etc is something @jmervine needs to confirm on Heroku.

Agreed. Again, this was more of a byproduct of C&P. The exact verbiage can be tweaked as needed. I just felt there should be some sort of documentation regarding the API.

scripts/api.js
XhmikosR: It doesn't allow symlinks IIRC. Maybe @jmervine has more info on this, since I don't recall the exact details rn.

scripts/api.js
XhmikosR: And explicitly disable the rule for Array destructuring :P

scripts/api.js
markcarver: FWIW, this really only matters locally when running the script to generate the new `_api.yml` file. This wouldn't actually be executed on AWS.

scripts/api.js
markcarver: The `prefer-destructuring` is a preference to enforce the usage of destructuring, not disallow them entirely. I really don't care either way, I just didn't see any lint errors when writing this and thought it was "ok". If you want me to change this I will; I'm not going to nitpick this.

markcarver:
I went ahead an moved some of the code into `helpers.js` in anticipation of use throughout the rest of the code. This way it'll be easier to transition from the old config to the new API.

markcarver:
Pushed some minor documentation changes

markcarver:
Just added support for this back in https://www.drupal.org/project/bootstrap/issues/3031415#comment-12994325.

It's using what will be the [new public endpoint URLs](http://cgit.drupalcode.org/bootstrap/tree/src/Plugin/Provider/BootstrapCdn.php) according to this PR, but as this isn't yet merged/deployed, it's not actually functional yet.

I had been using the express endpoints to test the API locally with Drupal, so decided since the code's there why not just commit it in anticipation of this.

There's proper error handling/messaging for when an API is unavilable/fails, so it's not that big of a deal. It would, though, be really nice to get this live before I have to make a release in the next week or so.

If it turns out that I need to release before this is ready, I can easily and temporarily disable it until this is. Just haven't heard much after the initial review.

Seriously, not trying to put on pressure, just trying to keep communication open so y'all know where I'm at 😄

Cheers!

markcarver:
Hello?

markcarver:
4 months now and still no response...

lib/helpers.js
```diff
@@ -64,7 +66,108 @@ function generateDataJson() {
...
+
+    raw(name, endpoint = config.apiVersion) {
+        return (config.api[endpoint] && config.api[endpoint][name]) || null;
+    },
+
+    send(res, data, statusCode = null) {
+        if (!statusCode) {
+            statusCode = data.status ? data.status : 200;
+        }
+        return res.status(statusCode).json(data);
```
Johann-S: useless return here

lib/helpers.js
```diff
@@ -64,7 +66,108 @@ function generateDataJson() {
     return data;
 }
 
+/**
+ * Helper methods used to handle the API.
+ */
+const api = {
+
```
Johann-S: trailing line

routes/api.js
```diff
@@ -0,0 +1,35 @@
...
+// API catch-all bad request.
+router.get('*', (req, res) => {
+    const data = helpers.api.badRequest();
+
+    helpers.api.send(res, data);
+});
+
+module.exports = router;
+
+// vim: ft=javascript sw=4 sts=4 et:
```
Johann-S: can you remove that please ?

scripts/api.js
```diff
@@ -0,0 +1,115 @@
...
+const yaml       = require('js-yaml');
+const config     = require('../config');
+const sri        = require('./sri.js');
+
+const configFile = config.getConfigPath('_api.yml');
+
+const semver = require('semver');
+
+function getDirectoryNames(source, ignore = []) {
+    return fs.readdirSync(source).map((name) => ignore.indexOf(name) === -1 && name).filter((name) => {
```
Johann-S: Can you indent a bit more your code here, for example:
```js
return fs.readdirSync(source)
  .map((name) => ignore.indexOf(name) === -1 && name)
  .filter((name) => {});
```

better for the readability

scripts/api.js
```diff
@@ -0,0 +1,115 @@
...
+
+// @todo Remove twitter-bootstrap once files are moved to the proper directory.
+const packages = getDirectoryNames(cdnPath, ['css', 'js', 'twitter-bootstrap']);
+const endpoints = {};
+
+endpoints.v1 = generateApiV1(packages);
+
+const generated = yaml.dump({ api: endpoints }, { lineWidth: -1 });
+
+const header = [
```
Johann-S: I think doing that:
```js
const header =
  `# THIS FILE IS GENERATED AUTOMATICALLY. DO NOT MODIFY THIS FILE DIRECTLY.
# Last generated: ${(new Date()).toString()}`;
```

should do the trick here, without using an array 🤔

routes/api.js
markcarver: All the JS files have this, I was just following existing conventions.

markcarver:
@Johann-S, I have addressed your feedback.

markcarver:
It's clear this isn't going anywhere. The one downside to FOSS: great work being done and ignored.

jimaek:
@markhalliwell I know its an old PR but I wanted to let you know that bootstrapcdn is now part of jsDelivr and we are in the process of migrating everything to a new system with a real API backed by https://github.com/jsdelivr/data.jsdelivr.com. All feedback is appreciated.

PR #1290, opened 2019-03-07
Updates

XhmikosR:


XhmikosR:
No, but I thought we'd use it in its original size. Not sure why I went with 150px width in the first place, but it looks good bigger too.

PR #1310, opened 2019-03-28
Switch to CircleCI

XhmikosR:
We need to make sure we don't rely on Travis for hooks, notifications etc.


jdorfman:
Has Travis gotten that bad since the acquisition?

XhmikosR:
I think the cause for the recent slowdowns was some incidents they had, but after trying CircleCI, it's just so much faster.

I still need to tweak things to reuse some stuff, but as you can see, it's very fast.

XhmikosR:
@jmervine if you want to add Slack notifications there's a Slack CircleCI app. Just make sure you remove the Travis app.

In the repo webhooks everything Travis-related I can remove it myself. The same for "Integrations & services".

jmervine:
What’s the motivation behind this switch?

XhmikosR:
The motivation is that CircleCI has a lot more capabilities and seems a lot
faster than Travis when the deps are cached. Also I'm not sure about Travis
stability; they had an incident recently which took 1-2 days to sort, but
that might be a coincident or not.

The only thing we lose is Windows testing but I think they plan to add this
later.

PR #1319, opened 2019-04-13
Update devDependencies

XhmikosR:


XhmikosR:
I wait for jQuery v3.4.1 because v3.4.0 has a bug that affects Bootstrap.

XhmikosR:
I stopped waiting for a new jQuery release. We'll update it when I confirm it works fine upstream.

PR #1326, opened 2019-05-03
Add Bootlint 0.16.6. and update deps

XhmikosR:
Needs the files on the origin, including the `/latest/` endpoint.


XhmikosR:
@jmervine it seems the files aren't uploaded yet? Don't forget the `/latest/` endpoint too.

jmervine:
@XhmikosR Uploaded. Restarting CI now. Might need to clear the cache for latest to succeed.

PR #1334, opened 2019-06-13
Use `codefund.io` for ad render

coderberry:
We are now using codefund.io instead of codefund.app for rendering ethical ads. I also added the new CDN URL that we will be migrating to. Finally, I updated your link to CodeFund to use your referral link.

### DIFF
--- a/config/helmet-csp.js
+++ b/config/helmet-csp.js
@@ -46,6 +46,7 @@ const CSP = {
         'codefund.app',
         'cdn2.codefund.app',
         'codefund.io',
+        'cdn.codefund.io',
         'launchbit.com',
         'www.launchbit.com',
         'www.ziprecruiter.com'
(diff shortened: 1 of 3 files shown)


jdorfman:
Weird, @XhmikosR added `codefund.io` this morning but still not working. I guess it needs `cdn.codefund.io`? 

https://github.com/MaxCDN/bootstrapcdn/pull/1333/files#diff-41812f0e88af3079ba60fefbc57498f8R23

XhmikosR:
When I tried it, it worked for me yeah.

Anyway, thanks @coderberry, but you guys should stop changing this stuff. :)

coderberry:
image

@XhmikosR do you know if it's cacheing the old policy? I've tried both Chrome and Safari

XhmikosR:
This isn't deployed to prod yet.

XhmikosR:
Should be in prod now.

PR #1338, opened 2019-06-16
Minor tweaks

XhmikosR:
WIP, to check the ads loading.


XhmikosR:
Unfortunately it seems I can't really test it in the PR preview because CodeFund says it doesn't have an ad. Must be some whitelisting they do based on the domain.

PR #1350, opened 2019-07-22
Switch to Node.js 12.x.

XhmikosR:


docker-compose.yml
```diff
@@ -1,5 +1,5 @@
 web:
-  image: jmervine/herokudev-node:10
+  image: jmervine/herokudev-node:12
```
XhmikosR: @jmervine: not sure if this image exists yet.

XhmikosR: Ping @jmervine

jmervine:
I'm not either, however given that I'm probably the only person that uses
this, you can merge away and I'll verify once I can.

PR #1362, opened 2019-09-20
Add lgtm.com support.

XhmikosR:
@jmervine can you enable the lgtm GitHub App? Now it's using the hooks which is deprecated.


XhmikosR:
OK, I think I installed the app, so we are all set. Now I want to see if the cdn files are excluded properly.

XhmikosR:
@jmervine what's your take on this? It seems lgtm is a little slow, so, I'm not sure if we should proceed. Could be I'm doing something wrong, but I'd expect it to be faster with excluding the cdn files...

PR #1365, opened 2019-09-30
Split Actions.

XhmikosR:


XhmikosR:
This seems to work well. The only thing I don't like is that when we make a PR ourselves we run things twice. This is something I haven't been able to sort out yet.

Note that current Actions don't have caching, so this will be even faster when that's implemented.

PR #1366, opened 2019-10-01
test/test_helpers.js: use `path.extname` to get the extension

XhmikosR:


test/test_helpers.js
```diff
@@ -22,14 +23,7 @@ let server = {};
...
 function getExtension(str) {
-    // use two enclosing parts; one for the dot (.)
-    // and one for the extension itself.
-    // So, the result we want is the third Array element,
-    // since the first one is the whole match, the second one
-    // returns the first captured match, etc.
-    const match = str.match(/(\.)([a-zA-Z0-9]+)$/);
-
-    return match && match[2];
+    return path.extname(str).slice(1);
```
XhmikosR: Note that str is a URL

XhmikosR: I'm still hesitant about this because str is a URL...

jmervine: What will it return if there's no extension? Generally I think it's okay that it's a URL if (a) it returns `nil` or `undefined` when there's no extension or (b) we can be certain that all `str` values passed contain extensions.

XhmikosR: An empty string is returned if there's no extension but then our tests will fail later anyway.

I will rethink about this.

PR #1372, opened 2019-10-08
Add `sr-only` for "Toggle Dropdown" buttons

XhmikosR:


XhmikosR:
@patrickhlauke: can you have a quick look please? Thanks!

XhmikosR:
@patrickhlauke: friendly ping :) Also, maybe we could just use `aria-label` instead? Not sure what's better.

patrickhlauke:
can't see anything on the preview, but working out what i assume the `.pug` things render to, it looks fine to me. `aria-label` is also an option, sure.

XhmikosR:
@patrickhlauke is there any preferable way? The preview should be up now.

patrickhlauke:
they're equivalent, as they lead to the same end result.

btw the "Old Versions" and "Resources" in the top navigation could also do with this addition.

XhmikosR:
Thanks for the replies! I think `aria-label` is the "simplest", i.e. no need to add a new element. I'm gonna experiment with it first.

About "Old Versions" and "Resources" are you sure it's needed? I mean those have text in them.

patrickhlauke:
> About "Old Versions" and "Resources" are you sure it's needed? I mean those have text in them.

ah, good point, i was probably being a bit overzealous. no you're right, they'll announce as "Old versions - collapsed" / "Old versions - expanded" ... which should be clear enough indication. yeah, forget that i mentioned them.

XhmikosR:
@patrickhlauke can you check one last time please? I switched to `aria-label='Toggle Dropdown'`

patrickhlauke:
LGTM

completely unrelated: maybe add `readonly` attribute to the text inputs? there's no point keeping them editable (and with readonly, they can still be focused/have the copy behavior)

XhmikosR:
Thanks for confirming!

Feel free to make a PR, otherwise I'll make an issue to track this.

PR #1379, opened 2019-10-14
Add `readonly` in text `input`s.

### LINKED ISSUE #1377: Investigate adding `readonly` attribute to the text inputs
https://github.com/MaxCDN/bootstrapcdn/pull/1372#issuecomment-541656991

XhmikosR:
Fixes #1377 

@patrickhlauke: LGTY?

Preview: <https://bootstrapcdn-dev-pr-1379.herokuapp.com/>

Also, @jmervine @jdorfman if you like the change; it changes the look, but I believe this is technically more correct. So, since copying, focus etc works, I think this should be fine.


patrickhlauke:
this looks good yes :shipit:

PR #1378, opened 2019-10-14
Tweak Bootswatch views.

XhmikosR:
* add title in thumbnails
* add `aria-label` in `input`s

I need some grammar proofing. Should it be `Celurian theme's thumbnail` or something else?

PR #1380, opened 2019-10-14
Production release

XhmikosR:
I might wait a bit for #1379 #1378


jdorfman:
LGTM, looks cleaner with just the ad, maybe it will get more clicks

PR #1404, opened 2019-11-06
Replace themes.

zoltanszogyenyi:
Replaced two themes that were no longer available with newer ones from Themesberg.


XhmikosR:
@jdorfman do we still need to keep the themes page?

zoltanszogyenyi:
Just did that and they all got skipped.

XhmikosR:
I definitely get some savings here.

jdorfman:
@XhmikosR please merge this. We can evaluate the need for a themes page in the future. =)

XhmikosR:
It will have to wait until tomorrow morning my time.

zoltanszogyenyi:
@XhmikosR I optimized the images with setting the max variable to 80 and managed to save some more.

XhmikosR:
Thanks but I'm not a fan of lossy optimization for jpeg. Especially in
images like these.

I will do it myself tomorrow.

jdorfman:
Thanks @XhmikosR.

PR #1414, opened 2019-11-21
Production release

XhmikosR:
>160 additions and 1,725 deletions 🎉

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -74,9 +74,9 @@
       }
     },
     "@babel/parser": {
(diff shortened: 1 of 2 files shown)


XhmikosR:
Shouldn't affect production but still, so happy to see such reduction in deps :)

XhmikosR:
I don't use it for sure. Where/how do we use it? I don't expect any JS errors TBH, mostly CSP errors but that's another story.

So, yeah, maybe we should remove it and also from Heroku.

PR #1417, opened 2019-11-26
Add bootstrap v4.4.0

XhmikosR:


XhmikosR:
@jmervine ping me when it's live

jdorfman:
@XhmikosR files are live, having some issues with the headers. @jmervine we should document this process so all three of us can do it going forward. =)

XhmikosR:
@jmervine can you fix the content types please?

XhmikosR:
Thanks!

PR #1420, opened 2019-11-27
Add new versions

XhmikosR:
@jmervine please ping me when it's live


XhmikosR:
Actually, let's wait a bit, we are going to have to release a v4.4.1 which I'm not sure it it affects Bootswatch or not yet. I need to check which Sass implementation they are using. Regardless, I'll push in this PR any further updates and ping you to upload the files.

jdorfman:
@XhmikosR I would prefer it if you worked with @jmervine to upload (and everything else) so we are fully redundant. We shouldn't have to ping Josh if he is on holiday or is just not reachable.

XhmikosR:
@jdorfman I can't know if and when someone is on vacation or not reachable. Also, there's just one too many things a person can do.

jdorfman:
@XhmikosR true that. I will take this one and futures if needed. =)

PR #1462, opened 2020-02-20
scripts: change permissions to 0644.

XhmikosR:


XhmikosR:
@jmervine this doesn't change the real permissions, only the permissions git has tracked. https://git-scm.com/docs/git-config#Documentation/git-config.txt-corefileMode

PR #1512, opened 2020-04-13
Tweak theme query

XhmikosR:
Use `Number.parseInt` since we expect this to be an integer.

@jmervine I think this is a better solution, WDYT?


XhmikosR:
I'm putting this on hold because I found another issue with that code. I need to first deduplicate the code and then fix it along with the tests.

PR #1531, opened 2020-04-28
Add CodeQL Analysis workflow

jhutchings1:
Adds a GitHub Actions workflow that runs CodeQL on every push, and on a daily schedule.

Code scanning looks for vulnerabilities, such as XSS, SQL injection, etc., in your code. If it finds any new vulnerabilities it surfaces them in the PR as check annotations, and blocks the build until they’re fixed or marked as false positives. If it finds any on the repo’s default branch it displays them in the [security tab](../security/code-scanning).

For now you also need to be feature flagged individually to see results in the security tab (as well as having write permission on this repo) - if you drop an email to jhutchings1@github.com I can get anyone you need added. 

Finally, this is an early access program that has not been released yet, so please don't share before May 6th when we’re unveiling it at GitHub Satellite.

Cc: @greysteil


jdorfman:
thanks @jhutchings1 

@XhmikosR I think we should merge this. Thoughts?

XhmikosR:
I'm in touch with @jhutchings1 since a few days now, I'll take care of this when it's ready :)

XhmikosR:
@greysteil: is there any way we can ignore certain folders? For example, we don't need to scan the `cdn` folder. I do the same for lgtm, although LGTM is a little slow regardless: https://github.com/MaxCDN/bootstrapcdn/blob/develop/.lgtm.yml

BTW, in theory, now we should see some XSS errors because some of the CDN files do have some vulnerabilities. I'm waiting for the run to finish to check out the results 🙂

greysteil:
Sure is - there are docs [here](https://github.com/github/codeql-action#configuration).

PR #1532, opened 2020-04-29
codeql.yml: ignore the `cdn` folder

XhmikosR:


XhmikosR:
@greysteil: can you check the changes in this PR please?

I'm not sure if we should run the action on `pull_request` too. But I don't see the action firing.

greysteil:
Hmmm, it should be firing. Let me look into that.

On the config, looks like we have some work to do to improve the docs. I've added a commit to yours in [this PR](https://github.com/MaxCDN/bootstrapcdn/pull/1533) that will ignore that folder.

greysteil:
Looks like it's firing OK now - not sure what happened there. Will get the team to look into `be2081c`.

XhmikosR:
Thanks! I rebased and cleaned up the patches, let's see if this speeds up the scan 🙂

XhmikosR:
Ok, so far the first build on develop seems to keep going. I guess it's normal because the `cdn` folder does have so many files? https://github.com/MaxCDN/bootstrapcdn/runs/629088675

This run here seems to proceed faster but it still seems to be slow: https://github.com/MaxCDN/bootstrapcdn/pull/1532/checks?check_run_id=629212651. Maybe this will be faster after this patch here is merged so that the comparison sees there are no new files?

greysteil:
Nice! I also figured out what happened with the missing scans - it looks like they don't show up on the commit if a [workflow file is invalid](https://github.com/MaxCDN/bootstrapcdn/actions/runs/91011017). That's not a great UI - I'll tell the Actions team.

Finally, you were right to add `pull_request` as a trigger - without it, you won't get the CodeQL can on PRs from external contributors. I'm going to change our default template to include it, too.

XhmikosR:
I'll  wait for the current runs to finish first and then I'll push the `pull_request` change here :)

BTW I submitted a small PR upstream, no functional changes just some tweaks https://github.com/github/codeql-action/pull/3

XhmikosR:
Hehe, it works! 😛

https://github.com/MaxCDN/bootstrapcdn/security/code-scanning

XhmikosR:
BTW there are definitely more issues that don't seem to be reported there. I mean, last year we fixed an XSS in bootstrap. Shouldn't that be reported?

greysteil:
Hurrah! I can't see those alerts as this  is your repo, but do feel free to email any feedback!

On speed, analysis should be the same speed as LGTM. We're always working to speed it up, but in some places that's a hard problems with slow progress. If the performance is meaningfully different from LGTM we should definitely be looking at that, though!

XhmikosR:
We only seem to be getting 2 rules triggered in the `cdn` folder: `js/incomplete-sanitization` and `js/unsafe-jquery-plugin`. I'd still expect the tool to flag the bootstrap XSS vulnerabilities, see https://github.com/twbs/bootstrap/releases/tag/v4.3.1 and https://github.com/twbs/bootstrap/releases/tag/v3.4.1.

This repo turned out as a playground for the tool, I hope you don't mind bringing up the issues here 🙂

greysteil:
Don't mind at all - thanks for experimenting with it and for the feedback!

I'm not sure why we're not flagging those XSS vulnerabilities, but can get someone from the JS languages team to take a look. I'm also asking someone from code scanning with more knowledge of performance to take a look here and get back to you with more details there :octocat:

XhmikosR:
Actually it seems the XSS is reported, but it's inside the `js/unsafe-jquery-plugin` rule which I didn't quite get it by the rule's name and I had to click on a warning to see what it's flagging:

Untitled

Anyway, I'll merge this PR after the check is done and see if the warnings go away.

But it's definitely slower than LGTM the current run has taken 13m and is still going :/

XhmikosR:
@greysteil it seems the warnings are not removed even though the offending files are excluded on the latest develop?

greysteil:
Team are looking into it now, but it looks like the `cdn` path isn't getting excluded for some reason. Also  explains the speed!

XhmikosR:
Ah, makes sense! Let me know how it goes :)

greysteil:
Will be fixed by https://github.com/github/codeql-action/pull/5. 😬🤣

XhmikosR:
Too bad we can't rerun a successful run :/

greysteil:
Or trigger an action run on demand. The Actions team are working on the latter :-)

XhmikosR:
Yup, this is something I've been missing since the August 2019 launch 🙁

OK, so AFAICT, we need the ignore docs to be updated or even better, make the action accept a `paths-ignore` key. Unless this was fixed with the latest upstream patches, in which case I'd rather remove the config file from the repo to keep things clean 🙂

greysteil:
Still going to need the config file I'm afraid - the upstream change just fixes it so the `paths-ignore` key is used even if no `queries` array was provided in the config file (the original behaviour looks like it was a copy-paste error).

XhmikosR:
Maybe this could be an improvement for later?

PR #1534, opened 2020-04-29
Production release

XhmikosR:


XhmikosR:
Waiting until we stabilize the CodeQL integration and #1535.

PR #1550, opened 2020-05-15
Add Edubirdie to footer as sponsor

jdorfman:


jdorfman:
They came back on as a sponsor https://opencollective.com/getbootstrapcdn/transactions

XhmikosR:
@jdorfman I don't monitor OC, so it'd be better if there was some kind of communication.

PR #1564, opened 2020-05-26
S3 upload automation

jmervine:
# Context

After a discussion with @jdorfman, we'd like to leverage using Github as the source of truth for automated syncing of cdn files to S3. This is to ensure prompt release of new files, as well as remove any question about state.

# Solution

Leverage existing ops automation, from https://github.com/jmervine/bootstrapcdn-ops. Wrap it in a Docker image for execution and use CI to automate via Github Actions.

# Testing

1. Docker image should run if configured properly.
1. CI should automatically run and sync when this branch is deployed, either at merge time, or actual deployment time, depending on how CI is implemented.


XhmikosR:
I can't actually comment on how this works, but why don't we use an existent action?

jmervine:
Not sure what "an existent action" is.

XhmikosR:
https://github.com/marketplace?type=actions&query=s3 or another search should give more insight.

.github/workflows/updater.yml
```diff
@@ -0,0 +1,15 @@
+name: Update S3
+
+on:
+  push:
+    branches:
+      - master
```
XhmikosR, line 6: Should probably be develop? `master` is only for production, right?

jmervine: Yes, this should only be run when deploying to production.

jimaek:
I believe this is not necessary anymore, too bad because its a cool solution

PR #1665, opened 2021-01-22
CDN additions

XhmikosR:


XhmikosR:
@jmervine could you please add the files from this branch to the origin? So that we wrap things up and call it a day.

PR #1675, opened 2021-03-07
Adapt functional tests to the recent CDN changes

### LINKED ISSUE #750: Enable HSTS protection for CDN assets.
Enabling HTTP Strict Transport Security will afford better protection to CDN traffic. 

Currently assets can be served over HTTP or HTTPS from the CDN, there is no redirection from HTTP to HTTPS. By enabling HSTS, capable clients will automatically upgrade to using HTTPS whilst those that are not capable, or may not be capable, will remain on HTTP.  

This can be done in the following ways:

1) Inject the Strict-Transport-Security header when serving assets. 
2) HSTS preload the CDN domain. 
3) A combination of 1 and 2.

XhmikosR:
TODO:

- [x] Streamline the headers we check
- [x] Enable compression tests

Closes #750


XhmikosR:
@jimaek it seems that the CDN tests are very slow after the CDN change. Not sure if it's related to the compression or not hitting the CDN cache? Previously they only took something like 30-35s when we hit the cache.

jimaek:
It's possible they didn't hit the cache for some older versions as we only started serving traffic a few days ago. Did a re-run help? I think they should hit a warm cache after the first test. If not do you have any kind of details like URLs and HTTP headers?

XhmikosR:
So, apart from the speed, it seems some headers are not present either like x-content-type-options and some others differ.

Please check the run and let me know if I should adapt the tests: https://github.com/MaxCDN/bootstrapcdn/pull/1675/checks?check_run_id=2050460079

Now it'd be a good change to streamline the tests.

PS. the content-type headers I still think should match what we had before, even though browsers are smart, I'd like to be spec-compliant.

jimaek:
Yeah some headers were removed. As for content we use `content-type: application/javascript; charset=utf-8`
If its different somewhere then its an older version that needs purging from the CDN

jimaek:
I set the content header based on the wiki https://en.wikipedia.org/wiki/Media_type
I guess I could change it but I believe most people use the application version.

Also I dont think it makes sense to add nosniff on a public CDN. They can sniff all they want :)

jimaek:
All tests now pass locally for me. I just made a few commits to help out with this PR. 
@XhmikosR Take it from here and let me know if there is anything needed

XhmikosR:
> Also I dont think it makes sense to add nosniff on a public CDN. They can sniff all they want :)

They shouldn't sniff **at all**, that's the point. It's a security header.

test/functional_test.js
```diff
@@ -18,31 +15,17 @@ const responses = {};
 
 // Expects header names to be lowercase in this object.
 const expectedHeaders = {
-    'accept-ranges': 'bytes',
     'access-control-allow-origin': '*',
-    'cache-control': 'public, max-age=31536000',
-    'connection': 'Keep-Alive',
-    'content-length': '',
+    'cache-control': 'public, max-age=31919000',
```
XhmikosR, line 19: How is this number computed? The previous one was 1 year.

jimaek: Not sure, its set by the CDN when choosing "1 year" as caching time in the UI

XhmikosR:
There are file types served uncompressed, please check the failures.

Also, I notice these differences:

* accept-ranges is removed but it's already set
* 'connection': 'Keep-Alive', it's not set but I wonder if it makes any difference?
* content-length should be set if we test for compression, right?
* x-cache is removed but isn't it always set? Just note that `''` means just check if the header is present and not its value
* cdn-cache: I don't see this header when using curl, although tests seem to pass

XhmikosR:
I wonder if the slowness is due to the missing keep-alive header...

jimaek:
>There are file types served uncompressed, please check the failures.

Looks like we need to modify mime type for woff and woff2 files to be `font/x-woff`. Otherwise CF wont gzip it.
Do you agree?


>accept-ranges is removed but it's already set

I dont see it anywhere

>content-length should be set if we test for compression, right?

I dont think so

>x-cache is removed but isn't it always set? Just note that '' means just check if the header is present and not its value

x-cache was a stackpath specific header. CF doesnt use it

>cdn-cache: I don't see this header when using curl, although tests seem to pass

I can see it with curl. Try again, I purged the CDN cache

XhmikosR:
> Looks like we need to modify mime type for woff and woff2 files to be `font/x-woff`. Otherwise CF wont gzip it.
> Do you agree?

That's also non-standard IIRC but if that's needed to get it to work, sure :)

> I dont see it anywhere

```
curl -sILl -H "Accept-Encoding: gzip, deflate" https://stackpath.bootstrapcdn.com/bootstrap/4.5.2/css/bootstrap.min.css | findstr "Accept"
vary: Accept-Encoding
````

> I dont think so

I was seeing, just like accept-ranges but now I don't. Maybe purging was needed? That's why I was seeing the difference in headers, I think.

**Comment by XhmikosR** [comment]
Actually, can't you ask Cloudflare to compress the 2 font MIME types we use? I'm saying this, because they are used a lot, and I personally hadn't noticed until now. The origin of these are https://github.com/h5bp/server-configs-apache/blob/3c7db53404c23ba705ff56dda7096c82cc409e8c/h5bp/media_types/media_types.conf#L66-L67 so they are used a lot.

**Comment by jimaek** [comment]
It doesnt look like its something they do per account, its a global rule https://support.cloudflare.com/hc/en-us/articles/200168396-What-will-Cloudflare-compress-

You could try asking them on Twitter since everyone would benefit from this change

**Comment by jimaek** [comment]
Looks like there is a reason its not gzipped, they are already compressed. So I restored the previous mimes and your test should ignore woff* files

**Comment by XhmikosR** [comment]
> Looks like there is a reason its not gzipped, they are already compressed. So I restored the previous mimes and your test should ignore woff* files

Damn, missed that. Thanks!

I also added back no-sniff.

**Comment by XhmikosR** [comment]
The slowness is still present and unresolved, though. Either cache is not hit or something else is going on. I mean, even with a cache hit the tests are at least 3-4 times slower.

**Comment by XhmikosR** [comment]
You can see the runner's IP in the CI run. Not sure what else we could do to debug this. With SP, it was super fast even if there was no cache hit (but we were missing compression on the first 1-3 requests).

**Comment by XhmikosR** [comment]
For example the current CI run as we speak is at 5 minutes and only ~half of the tests are run.

**Comment by XhmikosR** [comment]
I'm also getting `font/x-woff` for the ATH POP.

File: test/functional_test.js [hunk 2]
```diff
@@ -18,50 +13,31 @@ const responses = {};
 
 // Expects header names to be lowercase in this object.
 const expectedHeaders = {
-    'accept-ranges': 'bytes',
     'access-control-allow-origin': '*',
-    'cache-control': 'public, max-age=31536000',
-    'connection': 'Keep-Alive',
```
XhmikosR, line 24: I do get this header here:

```
C:\Users\xmr\Desktop\bootstrapcdn>curl -ILl https://stackpath.bootstrapcdn.com/font-awesome/4.7.0/fonts/fontawesome-webfont.woff
HTTP/1.1 200 OK
Date: Wed, 10 Mar 2021 19:09:36 GMT
Content-Type: font/x-woff
Content-Length: 98024
Connection: keep-alive
Set-Cookie: __cfduid=dcc42ec6c8ee7885051fc2830c11525a31615403376; expires=Fri, 09-Apr-21 19:09:36 GMT; path=/; domain=.bootstrapcdn.com; HttpOnly; SameSite=Lax
CDN-PullZone: 252412
CDN-Uid: b1941f61-b576-4f40-80de-5677acb38f74
CDN-RequestCountryCode: IT
CDN-EdgeStorageId: 632
CDN-EdgeStorageId: 617
Last-Modified: Mon, 25 Jan 2021 22:04:55 GMT
CDN-CachedAt: 2021-03-09 18:47:57
CDN-RequestPullSuccess: True
CDN-RequestPullCode: 200
Cache-Control: public, max-age=31919000
timing-allow-origin: *
cross-origin-resource-policy: cross-origin
access-control-allow-origin: *
x-content-type-options: nosniff
CDN-RequestId: 63b430331fd7c16c40aff591a44fdee5
CDN-Cache: HIT
CF-Cache-Status: HIT
Age: 86777
Accept-Ranges: bytes
cf-request-id: 08bf2477dd00009e7991a21000000001
Expect-CT: max-age=604800, report-uri="https://report-uri.cloudflare.com/cdn-cgi/beacon/expect-ct"
Strict-Transport-Security: max-age=31536000; includeSubDomains; preload
Server: cloudflare
CF-RAY: 62ded69fcb679e79-ATH
alt-svc: h3-27=":443"; ma=86400, h3-28=":443"; ma=86400, h3-29=":443"; ma=86400
```

jimaek:
> I'm also getting `font/x-woff` for the ATH POP.

Should be fixed now.

> You can see the runner's IP in the CI run. Not sure what else we could do to debug this. With SP, it was super fast even if there was no cache hit (but we were missing compression on the first 1-3 requests).
> 
> For example the current CI run as we speak is at 5 minutes and only ~half of the tests are run.

Lets move this to a separate issue. Because as part of fixing the tests I had to purge our cache a few times which makes it even harder to debug this issue

XhmikosR:
I'm still not sure about Accept-Ranges which is present for me, and connection keep-alive which seems to be missing from the origin, otherwise Cloudflare should pass it along.

XhmikosR:
How about keep-alive? This could theoretically affect performance for http/1.1.

Anyway, I'll wait for this to pass and merge it tomorrow along with your other PR.

The slowness should definitely be investigated because it's a  pretty huge one. Uncached it looks like at least 800% slower. And even cached aprox 200% slower.

jimaek:
> How about keep-alive? This could theoretically affect performance for http/1.1. I have no idea what request (which is used for the tests) is using.

Looks like we only have <9% of traffic over HTTP1.1. I dont think its a problem

> Anyway, I'll wait for this to pass and merge it tomorrow along with your other PR.
> 
> The slowness should definitely be investigated because it's a pretty huge one. Uncached it looks like at least 800% slower. And even cached aprox 200% slower.

I agree, I will look into it

XhmikosR:
I'm still getting a few local failures (ATH POP)

```

C:\Users\xmr\Desktop\bootstrapcdn>npm run mocha:only-functional

> bootstrapcdn@0.0.1 mocha:only-functional C:\Users\xmr\Desktop\bootstrapcdn
> npm run mocha -- --reporter progress ./test/functional_test.js


> bootstrapcdn@0.0.1 mocha C:\Users\xmr\Desktop\bootstrapcdn
> cross-env NODE_ENV=test ENABLE_CRAWLING=true mocha "--reporter" "progress" "./test/functional_test.js"


  [▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬]

  19631 passing (1m)
  4 failing

  1) functional
       cdn/**/*.*
         https://stackpath.bootstrapcdn.com/font-awesome/4.3.0/fonts/fontawesome-webfont.woff2
           does NOT have content-encoding set:
     AssertionError [ERR_ASSERTION]: Expected values to be strictly equal:
+ actual - expected

+ 'gzip'
- undefined
      at Context.<anonymous> (test\functional_test.js:114:20)
      at processImmediate (internal/timers.js:461:21)

  2) functional
       cdn/**/*.*
         https://stackpath.bootstrapcdn.com/font-awesome/4.3.0/fonts/fontawesome-webfont.woff2
           has content-type:

      AssertionError [ERR_ASSERTION]: Invalid "content-type" for "woff2", expects "font/woff2" but got "font/x-woff"
      + expected - actual

      -font/x-woff
      +font/woff2

      at assertContentType (test\functional_test.js:124:12)
      at Context.<anonymous> (test\functional_test.js:290:21)
      at processImmediate (internal/timers.js:461:21)

  3) functional
       cdn/**/*.*
         https://stackpath.bootstrapcdn.com/font-awesome/4.5.0/fonts/fontawesome-webfont.woff2
           does NOT have content-encoding set:
     AssertionError [ERR_ASSERTION]: Expected values to be strictly equal:
+ actual - expected

+ 'gzip'
- undefined
      at Context.<anonymous> (test\functional_test.js:114:20)
      at processImmediate (internal/timers.js:461:21)

  4) functional
       cdn/**/*.*
         https://stackpath.bootstrapcdn.com/font-awesome/4.5.0/fonts/fontawesome-webfont.woff2
           has content-type:

      AssertionError [ERR_ASSERTION]: Invalid "content-type" for "woff2", expects "font/woff2" but got "font/x-woff"
      + expected - actual

      -font/x-woff
      +font/woff2

      at assertContentType (test\functional_test.js:124:12)
      at Context.<anonymous> (test\functional_test.js:290:21)
      at processImmediate (internal/timers.js:461:21)

```

jimaek:
I guess the purge didnt propagate yet. I removed the changes. I think it should be safe to merge

XhmikosR:
@jimaek I'm still getting the above 4 test failures here on ATH POP.

jimaek:
Those URLs works fine for me but I purged them again

XhmikosR:
Still the same... Maybe contact Coudflare? I'm hitting the ATH POP. For example:

```
C:\Users\XhmikosR\Desktop>curl -ILl https://stackpath.bootstrapcdn.com/font-awesome/4.5.0/fonts/fontawesome-webfont.woff2
HTTP/2 200
date: Thu, 11 Mar 2021 10:54:35 GMT
content-type: font/x-woff
content-length: 66624
set-cookie: __cfduid=dfb93c0eb0e4bbc594e538138fb6a45ac1615460075; expires=Sat, 10-Apr-21 10:54:35 GMT; path=/; domain=.bootstrapcdn.com; HttpOnly; SameSite=Lax
cdn-pullzone: 252412
cdn-uid: b1941f61-b576-4f40-80de-5677acb38f74
cdn-requestcountrycode: IT
cdn-edgestorageid: 722
cdn-edgestorageid: 617
cdn-edgestorageid: 617
cdn-edgestorageid: 617
cdn-edgestorageid: 617
last-modified: Mon, 25 Jan 2021 22:04:54 GMT
cdn-cachedat: 2021-03-10 20:26:22
cdn-requestpullsuccess: True
cdn-requestpullcode: 200
cache-control: public, max-age=31919000
timing-allow-origin: *
cross-origin-resource-policy: cross-origin
access-control-allow-origin: *
x-content-type-options: nosniff
cdn-requestid: 4e132af6f1b1447346ca070d01016c80
cdn-cache: HIT
cf-cache-status: HIT
age: 74
accept-ranges: bytes
cf-request-id: 08c285a1810000fd6a41a88000000001
expect-ct: max-age=604800, report-uri="https://report-uri.cloudflare.com/cdn-cgi/beacon/expect-ct"
strict-transport-security: max-age=31536000; includeSubDomains; preload
server: cloudflare
cf-ray: 62e43ee26ac3fd6a-ATH
alt-svc: h3-27=":443"; ma=86400, h3-28=":443"; ma=86400, h3-29=":443"; ma=86400
```

Plus the above 3 files

PR #1680, opened 2021-03-11
Remove wrong bootstrap.min.css files from bootswatch.

XhmikosR:
These were HTML files...

@jimaek these should be immediately removed from the CDN...


jimaek:
I will remove them but the damage is already done, I only assume all potential users noticed the problem and simply didnt use these files at all.

jimaek (approved):

PR #1692, opened 2021-03-20
Prod sync

jimaek:
Will keep this open until we are ready to deploy something. Testing heroku preview at the same time as well.


XhmikosR:
Heroku would probably need to be reconnected (I had some rights myself, but not full admin rights and I'm not sure what's the case after the repo move). Better ask Joshua about Heroku.

PR #1695, opened 2021-03-22
Cdn links

disko998:


disko998:
@jimaek

scripts/integrity1.js
```diff
@@ -0,0 +1,100 @@
+#!/usr/bin/env node
```
jimaek: I think this file can be safely deleted? Same for all files under `/cdn`

jimaek:
Fix the build and tests please. Under this PR we can also remove the `cdn/` directory I think. Please do that and fix any issues that come from that.

config/_files2.yml
```diff
@@ -0,0 +1,492 @@
+bootstrap:
```
jimaek, line 1: I dont think we need this file anymore, do we?

config/_files.yml
```diff
@@ -1,338 +1,323 @@
 bootstrap:
```
jimaek, line 1: this file can be removed I think, it was needed when the files were hosted in the repo itself. Now both cdn/ dir and this file can be removed I believe a new one can be generated during build?

config/_files2.yml
disko998: No this can be deleted

config/_files.yml
disko998: So this files is generated with a new generation script that creates cdn from jsDeliver api. And then later app reads from the file to generate sri and display links on the website.

jimaek: Yes so we can remove it from the repo and run it during the build process and generate the file then?

disko998: Yes, I just need to make a small change because the script now requires this file to exist before running the script, but I'll change that to create it on the fly.

XhmikosR:
I'd rather do this myself @jimaek

.prettierrc.js
```diff
@@ -0,0 +1,7 @@
+module.exports = {
```
XhmikosR, line 1: Why would you use prettier... prettier is just for formatting. And xo already supports it out of the box.

disko998: Is not a big deal man you can remove, because we are 2 contributors so I just wanted to make sure we get the same code formatting but I didn't know about xo.

MartinKolarik:
Are we showing an old (v4) version of font-awesome on purpose? There's v5 but in a new package: https://www.jsdelivr.com/package/npm/@fortawesome/fontawesome-free

jimaek:
@disko998 Indeed, we should use latest v5 on the main page and maybe keep v4 on the legacy page.
Check that we correctly use latest versions on all other projects as well.

MartinKolarik:
@disko998 we're talking about this:

image

disko998:
> @disko998 we're talking about this:
> 
> ![image](https://user-images.githubusercontent.com/6192491/113185709-10681a00-9257-11eb-8908-d4384942ac61

Ik, i refer to a different issue @jimaek knows

MartinKolarik:
As I said, v5 is in a new package: https://www.jsdelivr.com/package/npm/@fortawesome/fontawesome-free

disko998:
> As I said, v5 is in a new package: https://www.jsdelivr.com/package/npm/@fortawesome/fontawesome-free

So we need to switch to that one?

disko998:
I was looking at the bootstrap doc like @jimaek said so they still use 4.7.0

MartinKolarik:
Yes.

jimaek:
> I was looking at the bootstrap doc like @jimaek said so they still use 4.7.0

I meant check the format and what files they use, not literally use the same versions. The goal of this migration is to always show the latest available versions for all projects that this site hosts

config/_files.yml
```diff
@@ -54,6 +54,7 @@ bootstrap:
     javascriptEsmSri: sha384-sKZy8g2KJhBTFCD6cIg8d4EifJxaa8c/iYIERdeKorHWhAgZgQOfqOKMe3xBqye1
```
jimaek, line 54: I think this file should be removed completely from the repo. For auto-updates to work we need to generate it during deployment. I dont see a reason to host it in the repo

disko998:
> > I was looking at the bootstrap doc like @jimaek said so they still use 4.7.0
> 
> I meant check the format and what files they use, not literally use the same versions. The goal of this migration is to always show the latest available versions for all projects that this site hosts

I mean that was the latest version for the package I used, how should I know that we need to use @fortawesome/fontawesome-free

config/_files.yml
disko998: I can put it in gitignore

disko998:
What is the command to fix these eslint errors, is annoying to go over them by hand?

config/_files.yml
jimaek: Do everything you need to delete it from the repo and make sure the tests run on the file that is generated on the fly

disko998:
@jimaek is there a way to fix these lint errors with puglint, not sure what is a cmd? Can't fix it with `eslint --fix` also someone mentions prettier works automatically but it doesn't seems so.

jimaek:
> @jimaek is there a way to fix these lint errors with puglint, not sure what is a cmd? Can't fix it with `eslint --fix` also someone mentions prettier works automatically but it doesn't seems so.

Maybe `xo --fix`?

XhmikosR:
Pug lint doesn't have an auto-fix command. You will need to fix the errors manually. Just don't use tabs and pay attention to indentation since pug needs it. But the errors here are in JS, so `npm run xo -- --fix` should do the trick.

MartinKolarik (approved):

PR #1696, opened 2021-03-24
Change CDN links to jsdelivr

XhmikosR:
@jimaek so the bootlint, bootstrap and bootstrap@5 devDependencies are only used in the scripts to generate the SRI hashes. Assuming this will change in the future, they can be removed later.

In this PR I added the missing files.

PS. the slowness is still present.

### DIFF
--- a/config/_files.yml
+++ b/config/_files.yml
@@ -1,338 +1,388 @@
 bootstrap:
+  - version: 5.0.0-beta3
+    stylesheet: https://cdn.jsdelivr.net/npm/bootstrap@5.0.0-beta3/dist/css/bootstrap.min.css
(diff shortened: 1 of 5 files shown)


XhmikosR:
@jimaek please disable Coveralls from commenting, I don't have the rights since I'm a simple external collaborator.

XhmikosR:
As for the functional test failures, they need to be fixed separately.

XhmikosR:
Actually, I'd better leave this to you to handle. Feel free to use any of this or go with #1695

jimaek:
Changing the files list is a simple solution but I dont think it applies to our use-case. Our goal is to automate everything.
So that file could be generated on the fly using the jsdelivr API as source. Otherwise we will be stuck making manual PRs for every Bootstrap release.

I will move some changes from here to the other PR. Feel free to comment there with potential fixes and improvements.

